PR #18, opened 2024-06-17
Move users table out of auth

ianmacartney:
Move the user creation/ update out of convex-auth, and provide a callback `onSignIn` that creates / updates the user record

----


xixixao:
Implemented via `createOrUpdateUser`

PR #32, opened 2024-07-12
allow overriding the audience

ianmacartney:
Allow non-convex audience so it can live side-by-side with a clerk provider (which afaik has to be convex at this point?)

----


xixixao:
This is not required, the applicationID is specific to the issuer. There can be multiple providers with the same applicationID.

To make sure I made an app that uses Clerk and Convex Auth on the same page (via two different clients):

Screenshot 2024-07-12 at 15 14 36

PR #39, opened 2024-07-19
Node10 resolution stubs and sourcemaps

thomasballinger:


xixixao:
Only `react` should need the stub files for React Native, the providers are used from Convex so they shouldn't need them?

thomasballinger:
React Native code imports types from code that imports from "convex/server", so I was thinking this might solve the issue for someone in Discord

PR #42, opened 2024-07-26
Fix broken links in api_reference.mdx

djbalin:
This PR fixes broken/dead links in the `api_reference.mdx` file found at the page for [API Reference](https://labs.convex.dev/auth/api_reference) in the docs. 

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #43, opened 2024-07-26
Fix broken links

mikewheaton:
This PR fixes some broken links in the documentation. I wasn't able to test it locally (running into a build error) but the fixes are simple and I've double-checked the URLs.


mikewheaton:
I should have checked the open PRs first, it looks like #42 includes similar changes for `api_reference.mdx`.

PR #44, opened 2024-07-30
[WIP] Add support for server-side authentication with Next.js App Router

xixixao:
Also removes refresh tokens and refresh token reuse detection for a simpler approach with bare session IDs.

TODO: 
- [x] Check `Origin` header before returning cookies
  -   ~[ ] And allow skipping the check for CORS support~ Not gonna support for now
- [x] Write docs on handling XSS
- [ ] Require the next version of Convex with web socket pausing for token fetching


xixixao:
Already merged on main.

PR #48, opened 2024-08-05
Callback for default user creation

xixixao:
I think this is good to add for now. My hope is that we'll find the time to implement a type-safe version of the library, and in that one we wouldn't need either of these callbacks.

Based on the discussion here: https://discord.com/channels/1019350475847499849/1270010190888177767/1270010190888177767

### DIFF
--- a/docs/pages/advanced.mdx
+++ b/docs/pages/advanced.mdx
@@ -106,6 +106,32 @@ use.
 
 {/* This is required when you configure [anonymous sign-in](/config/anonymous). */}
 
+## Writing additional data during authentication
+
+If you don't specify the
+[`createOrUpdateUser` callback](#controlling-user-creation-and-account-linking-behavior)
+you can still perform additional writes to the database with the
+[`afterUserCreationOrUpdate` callback](/api_reference/server#callbacksafterusercreationorupdate):
+
+```ts
+import GitHub from "@auth/core/providers/github";
+import Password from "@convex-dev/auth/providers/Password";
+import { DataModel } from "./_generated/dataModel";
+
+export const { auth, signIn, signOut, store } = convexAuth<DataModel>({
+  providers: [GitHub, Password],
+  callbacks: {
+    // `args` are the same the as for `createOrUpdateUser` but include `userId`
+    async afterUserCreationOrUpdate(ctx, { userId }) {
+      await ctx.db.insert("someTable", { userId, data: "some data" });
+    },
+  },
+});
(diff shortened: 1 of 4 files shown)


docs/pages/api_reference/server.mdx
```diff
@@ -248,7 +248,7 @@ or throws an error.
 
 <h3 class="nx-font-semibold nx-tracking-tight nx-text-slate-900 dark:nx-text-slate-100 nx-mt-8 nx-text-2xl">Defined in</h3>
 
-[src/server/implementation.ts:1344](https://github.com/get-convex/convex-auth/blob/main/src/server/implementation.ts#L1344)
+[src/server/implementation.ts:1350](https://github.com/get-convex/convex-auth/blob/main/src/server/implementation.ts#L1350)
```
ianmacartney, line 251: out of curiosity - are the line numbers here being updated automatically? I'd love to get helpers docs looking good like auth / ents, and want to know any tips & tricks

xixixao: Yeah, this is from `typedoc` and `typedoc-plugin-markdown`, which we use in main Convex docs as well.

docs/pages/advanced.mdx
```diff
@@ -106,6 +106,32 @@ use.
...
+```ts
+import GitHub from "@auth/core/providers/github";
+import Password from "@convex-dev/auth/providers/Password";
+import { DataModel } from "./_generated/dataModel";
+
+export const { auth, signIn, signOut, store } = convexAuth<DataModel>({
+  providers: [GitHub, Password],
+  callbacks: {
+    // `args` are the same the as for `createOrUpdateUser` but include `userId`
+    async afterUserCreationOrUpdate(ctx, { userId }) {
```
kstulgys, line 125: `afterUserCreationOrUpdate` => `afterUserCreatedOrUpdated` ?

erquhart:
How would type safety replace event hooks?

xixixao:
@erquhart the library would be more compositional, so you'd do something like (very much pseudo code):

```ts
import { createAccount, signIn } from "@convex-dev/auth/server"

export const passwordSignUp = mutation({
  args: {
    username: v.string(),
    password: v.string(),
  },
    const userId = await ctx.db.insert("users", {username: args.username});
    const account = createAccount(ctx, userId, {identifier: args.username, secret: args.password});
    return await signIn(userId);
  }
});

PR #56, opened 2024-08-16
Ian/swap index

ianmacartney:
swap the index order so you could list all the accounts for a given provider in the future

----


xixixao:
This the order we had before you told me to swap it: a02a8f69

ianmacartney:
Sorry for the trash - I still think it should be [userId, provider], but
[provider, providerId] seems more useful b/ca user can have many providers,
but a provider ID should only have one provider, right? And I'd assume you
might want to audit all the accounts for a provider for a migration or
deprecation, say, and you also might want to list all the providers for a
user

Issue #60, opened 2024-08-30
OAuth + Next.js not working in Safari


sshader:
Can reproduce with Github OAuth where we see "Invalid verifier" in `auth:store` when going through the flow, but the same flow works in Chrome.

e-hart:
Experiencing the same. Have reproduced with password and magic link providers in Safari on both macOS and iOS.

sshader:
Should be resolved in 0.0.63

PR #62, opened 2024-09-03
Don't use secure: true cookies for localhost

sshader:
See https://github.com/get-convex/convex-auth/issues/60 -- the alternative approach would be to make this a flag that's passed in + a note for Safari users to turn this on instead of doing this automatically.


thomasballinger (approved):
nice

PR #63, opened 2024-09-04
Add refresh token mutex + don't leak beforeunload listener

sshader:
I believe this fixes two bugs:
* In React Native (or any environment where `window.navigator.locks` is undefined), we can get re-entrant calls to `fetchAccessToken` where one call clears the old refresh token, and before it can set it to a new value, another call reads the intermediate state and believes we're in a logged out state. We have `browserMutex` wrapping this to prevent multiple tabs from stepping on each other, which also has the side effect of ensuring that we don't have re-entrant calls in browsers. This adds a promise-based mutex storing state on a global.
* We have a `beforeunload` listener to prevent closing the tab when we're in the middle of getting a new refresh token. However, if that fails for any reason, we end up with a `beforeunload` listener that we never clean up (I discovered this on accident by turning off my network in a tab with short lived JWTs and being annoyed that I could never close my tab).

The React Native issue I could repro originally by using a short lived JWT and backgrounding my app over the refresh window no longer reproduces (we initially hit an AuthError on reconnect since the token has expired, but then when we fetch the new token, the auth manager accepts it instead of mistakenly believing that we failed to acquire a new token).

I tested the `beforeunload` behavior by verifying that when we fail to store the refresh token the tab will still close after. I also tested that it fires and blocks the tab closing when we're in the middle of getting the refresh token (injected a sleep + some log lines so I could manually test).

### DIFF
--- a/src/react/client.tsx
+++ b/src/react/client.tsx
@@ -78,6 +78,8 @@ export function AuthProvider({
   );
   const { storageSet, storageGet, storageRemove, storageKey } =
     useNamespacedStorage(storage, storageNamespace);
+
+  const [isRefreshingToken, setIsRefreshingToken] = useState(false);
   const setToken = useCallback(
     async (
       args:
@@ -112,6 +114,27 @@ export function AuthProvider({
     },
     [storageSet, storageRemove],
   );
+
+  useEffect(() => {
+    const listener = async (e: Event) => {
+      if (isRefreshingToken) {
+        // There are 3 different ways to trigger this pop up so just try all of
+        // them.
+
+        e.preventDefault();
+        // This confirmation message doesn't actually appear in most modern
+        // browsers but we tried.
+        const confirmationMessage =
+          "Are you sure you want to leave? Your changes may not be saved.";
+        e.returnValue = true;
+        return confirmationMessage;
+      }
+    };
+    browserAddEventListener("beforeunload", listener);
+    return () => {
+      browserRemoveEventListener("beforeunload", listener);
+    };
+  });
 
   useEffect(() => {
     // We're listening for:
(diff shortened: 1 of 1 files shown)


src/react/client.tsx
```diff
@@ -226,19 +249,17 @@ export function AuthProvider({
           const refreshToken =
             (await storageGet(REFRESH_TOKEN_STORAGE_KEY)) ?? null;
           if (refreshToken !== null) {
+            setIsRefreshingToken(true);
```
sshader, line 252: I moved this up slightly because it feels like we probably shouldn't let you navigate away when we've cleared out your old token but before we've fetched the new one.

If `verifyCodeAndSetToken` fails for some reason, we'll treat this as if you've been logged out, but let you close the tab.

src/react/client.tsx
```diff
@@ -404,7 +425,81 @@ async function browserMutex<T>(
...
+    });
+  } else {
+    setMutexValue(key, {
+      ...mutex,
+      waiting: [...mutex.waiting, callback],
+    });
+  }
+}
+
+async function manualMutex<T>(
```
sshader, line 490: Open to suggestions on approach here. Over DMs we discussed creating an indefinitely long chain of promises, which also works but feels a little gross to me (I'm not sure if I have a chain of 100 resolved promises if they get collapsed and garbage collected).

I could also probably store this state on the auth provider instead of a global and get the same locking. Not sure if we dislike using a global.

src/react/client.tsx
```diff
@@ -113,6 +115,27 @@ export function AuthProvider({
...
+          "Are you sure you want to leave? Your changes may not be saved.";
+        e.returnValue = true;
+        return confirmationMessage;
+      }
+    };
+    browserAddEventListener("beforeunload", listener);
+    return () => {
+      browserRemoveEventListener("beforeunload", listener);
+    };
+  });
```
thomasballinger, line 137: no deps means this runs every render, which works but could be expensive if this provider isn't at the top

src/react/client.tsx
thomasballinger: Know what the goal of not allowing navigation away during this time was? Could there be queued mutations that wouldn't be covered by the existing warning?

src/react/client.tsx
thomasballinger: global makes sense, matches the other semantics

thomasballinger (approved):

src/react/client.tsx
sshader: My best guess is that we didn't want the user closing the tab at an inopportune moment to result in them being logged out when they come back?

PR #79, opened 2024-09-23
Gracefully handle errors thrown during /api/auth

sshader:
If talking to Convex fails for any reason (even though it shouldn't), I believe we end up sending a 500 with a generic Next.js error and fail to reset any state (i.e. clear cookies).

I've witnessed this hitting an error refreshing the session, and instead of clearing the cookies, we error and the client continues to have a useless refresh token in its cookies.

This also fixes an issue where we were requiring a valid auth token for `signIn`, when really we should just require one for `signOut`

### DIFF
--- a/src/nextjs/server/proxy.ts
+++ b/src/nextjs/server/proxy.ts
@@ -52,9 +52,16 @@ export async function proxyAuthActionToConvex(
 
   if (action === "auth:signIn") {
     let result: SignInAction["_returnType"];
+    // Do not require auth when refreshing tokens or validating a code since they
+    // are steps in the auth flow.
+    const fetchActionAuthOptions =
+      args.refreshToken !== undefined || args.params?.code !== undefined
+        ? {}
+        : { token };
     try {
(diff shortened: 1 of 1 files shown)


src/nextjs/server/proxy.ts
```diff
@@ -49,13 +49,27 @@ export async function proxyAuthActionToConvex(
...
-    url: options?.convexUrl,
-    token,
-  });
 
   if (action === "auth:signIn") {
-    const result = untypedResult as SignInAction["_returnType"];
+    let result: SignInAction["_returnType"];
+    // Do not require auth when refreshing tokens or validating a code since they
+    // are steps in the auth flow.
+    const fetchActionAuthOptions =
```
sshader, line 57: I added this just now to match how we use `authenticatedCall` and `unauthenticatedCall`. AFAICT the `signIn` function doesn't actually rely on being called as an authenticated call, but matching the existing behavior in the vanilla React integration seemed safer.

PR #98, opened 2024-10-16
Update password sign-up flow with a custom field

dowski:
Allows the user signing up to select a favorite color which is passed into the user creation flow and stored in the `users` table.

----


sshader (commented):
Shoot I had a whole comment drafted up and then forgot about it and closed the tab.

This is great! Let's make it as a separate provider though so we still have he basic email + password flow in the demo too.

The `profile` method encompassing both password requirements + user metadata is a little odd (but I don't really want to change it and deviate from Auth.js / what we already have too hard).

Thoughts on exposing something like `enforceDefaultPasswordRequirements` and updating our docs to show how to call it from a custom `profile`? With your change, technically it's no longer possible to have looser password constraints (e.g. if you want even shorter OTPs). I also realized our docs [here](https://labs.convex.dev/auth/config/passwords#customize-user-information) suggest something that in the current implementation completely skips any password validation.

dowski:
Thanks for the feedback! Updated to put the custom field stuff in its own demo. LMK if there's a better way to structure it since I'm just getting up to speed w/ React.

Re: password requirements, yeah, having something separate for that makes sense. Like you said though, I didn't want to blow away password requirements just because I wanted custom profile info. I'll come back to this with something like your `enforceDefaultPasswordRequirements` idea.

src/providers/Password.ts
```diff
@@ -107,9 +119,19 @@ export function Password<DataModel extends GenericDataModel>(
   return ConvexCredentials<DataModel>({
     id: "password",
     authorize: async (params, ctx) => {
+      const flow = params.flow as string;
+      if (flow === "signUp" || flow === "reset-verification") {
+        const password = (
```
sshader, line 124: nit: took me a bit to convince myself that this logic was equivalent to what we had before
```
const passwordToValidate = flow === "signUp" ? params.password : flow === "reset-verification" ? params.newPassword : null
```
makes a little more sense to me

test/convex/auth.ts
```diff
@@ -34,6 +35,29 @@ export const { auth, signIn, signOut, store } = convexAuth({
     TwilioVerify,
     TwilioOTP,
     Password,
+    // Sample password auth with a custom parameter provided during sign-up
+    // flow and custom password validation requirements (at least six chars
+    // with at least one number, upper and lower case chars).
+    Password<DataModel>({
+      id: "password-custom",
+      profile(params, _) {
```
sshader, line 43:
```suggestion
      profile(params, _ctx) {
```

test/src/auth/SignInFormPasswordAndCustomField.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { HexColorPicker } from "react-colorful";
+
+export function SignInFormPasswordAndCustomField() {
+  const [color, setColor] = useState<string | null>(null);
+  return (
+    <div className="max-w-[384px] mx-auto flex flex-col gap-4">
+      <h2 className="font-semibold text-2xl tracking-tight">
+        Sign in or create an account
+      </h2>
+      <SignInWithPassword
```
sshader, line 13: I suspect we could refactor this (don't love that the `customSignUp` depends on being rendered within a form), but this is fine for now.

One eventual goal is for the UI for the most common flows to be copy-paste-able (Clerk does a really good job of this with packaged `<SignIn/>` components), so we'll probably do a pass at all the components eventually.

src/providers/Password.ts
```diff
@@ -74,6 +74,18 @@ export interface PasswordConfig<DataModel extends GenericDataModel> {
...
+   *
+   * Otherwise the default validation is used (password is not empty and
+   * at least 8 characters in length).
+   *
+   * If the provided password is invalid, implementations must throw an Error.
+   *
+   * @param password the password supplied during "signUp" or
+   *                 "reset-verification" flows.
+   */
+  validate?: (password?: string) => void;
```
sshader, line 88: Why is the password optional? Also thoughts on `validatePassword` or `validatePasswordRequirements`? (as opposed to validating the login / validating that the password is correct)

dowski: Thanks, I like the longer/more clear name. I went with `validate` initially to match the other short API names - but clarity is good.

Also I made `password` optional b/c I guess I thought it might not be provided. But I guess in that case it can be modeled as empty string.

src/providers/Password.ts
dowski: Sure, that seems fine.

test/src/auth/SignInFormPasswordAndCustomField.tsx
dowski: Thanks for the context.

Since we want to offer customization of what's included in the sign up flow, having the `customSignUp` be part of the form makes sense to me. Do you have a different structure in mind though? Totally fine reworking it in a follow-up PR.

PR #100, opened 2024-10-17
Update reuse window implementation

sshader:
This updates our reuse detection for refresh tokens as follows:
* refresh tokens track their parent token
* the "active refresh token" is the most recently issued token that has never been used
* If a refresh token has never been used (its `firstUsedTime` is unset), issue a new token + update this tokens `firstUsedTime`
* If a refresh token is the parent of the active refresh token, return the active refresh token (regardless of when this refresh token was first used)
* If a refresh token has been used, but is within the allowed reuse token (10s), issue a new token
* Otherwise, this is an invalid reuse, so mark this token and all its descendants as usable (firstUsedTime 10+s ago)

This is more or less what supabase does for their refresh tokens with a pretty good write up of why [here](https://supabase.com/docs/guides/auth/sessions#what-is-refresh-token-reuse-detection-and-what-does-it-protect-from).

Some interesting effects:
* Once someone upgrades, they can't downgrade without migrating their data since I'm adding a column (we've done this before, and feels fine as long as we note it in our changelog)
* We don't clean up refresh tokens until we clean up the session (as opposed to previously we'd clean them up on each use). In the future, we'd perhaps want cron jobs to clean up stale sessions automatically.

Some other changes also in this PR:
* I added some more logging for my own sanity + ended up upgrading our version of convex-test so the new tests I wrote would work.

### DIFF
--- a/src/server/implementation/refreshTokens.ts
+++ b/src/server/implementation/refreshTokens.ts
@@ -1,6 +1,6 @@
 import { GenericId } from "convex/values";
 import { ConvexAuthConfig } from "../index.js";
-import { MutationCtx } from "./types.js";
+import { Doc, MutationCtx, QueryCtx } from "./types.js";
 import {
   LOG_LEVELS,
   REFRESH_TOKEN_DIVIDER,
@@ -9,11 +9,12 @@ import {
 } from "./utils.js";
 
 const DEFAULT_SESSION_INACTIVE_DURATION_MS = 1000 * 60 * 60 * 24 * 30; // 30 days
-const REFRESH_TOKEN_REUSE_WINDOW_MS = 10 * 1000; // 10 seconds
+export const REFRESH_TOKEN_REUSE_WINDOW_MS = 10 * 1000; // 10 seconds
 export async function createRefreshToken(
   ctx: MutationCtx,
-  sessionId: GenericId<"authSessions">,
   config: ConvexAuthConfig,
+  sessionId: GenericId<"authSessions">,
+  parentRefreshTokenId: GenericId<"authRefreshTokens"> | null,
 ) {
   const expirationTime =
     Date.now() +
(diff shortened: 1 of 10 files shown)


src/server/implementation/refreshTokens.ts
```diff
@@ -23,27 +24,89 @@ export async function createRefreshToken(
...
+    refreshTokenId: refreshTokenId as GenericId<"authRefreshTokens">,
+    sessionId: sessionId as GenericId<"authSessions">,
+  };
+};
+
+/**
+ * Mark all refresh tokens for a session as invalid. They will still be usable
+ * for 10 seconds (the length of the reuse window).
+ * @param ctx
+ * @param sessionId
```
dowski, line 56: Looks like this doesn't match the signature below (it uses `refreshToken`).

src/server/implementation/refreshTokens.ts
```diff
@@ -83,10 +136,25 @@ export async function validateRefreshToken(
...
-  }
   return { session, refreshTokenDoc };
 }
+/**
+ * The active refresh token is the most recently created refresh token that has
+ * never been used.
+ *
+ * @param ctx
+ * @param sessionId
+ * @returns
```
dowski, line 147: Did you mean to document a return type?

src/server/implementation/refreshTokens.ts
```diff
@@ -83,10 +136,25 @@ export async function validateRefreshToken(
...
+ */
+export async function loadActiveRefreshToken(
+  ctx: QueryCtx,
+  sessionId: GenericId<"authSessions">,
+) {
+  const refreshTokenDoc = await ctx.db
```
dowski, line 153: What about returning directly vs. creating the variable and returning it?

src/server/implementation/types.ts
```diff
@@ -83,7 +83,14 @@ export const authTables = {
     sessionId: v.id("authSessions"),
     expirationTime: v.number(),
     firstUsedTime: v.optional(v.number()),
-  }).index("sessionId", ["sessionId"]),
+    parentRefreshTokenId: v.optional(v.id("authRefreshTokens")),
```
dowski, line 86: Can you add some documentation above re: the parent relationship that's being tracked now?

dowski (approved):
Just a few minor comments, overall looks good.

src/server/implementation/refreshTokens.ts
sshader: I definitely had some debug logging here at one point and cleaned it up without also cleaning up the pointless variable 🙈

PR #101, opened 2024-10-17
Update to latest path-to-regexp

dowski:
This resolve a vulnerability present in the old version.

Required a slight API tweak as the pathToRegexp function now returns a different type.

https://github.com/advisories/GHSA-9wv6-86v2-598j

----

### DIFF
--- a/src/nextjs/server/routeMatcher.ts
+++ b/src/nextjs/server/routeMatcher.ts
@@ -52,7 +52,10 @@ export type RouteMatcherParam =
  * predefined routes that can be passed in as the first argument.
  *
  * You can use glob patterns to match multiple routes or a function to match against the request object.
- * Path patterns and regular expressions are supported, for example: `['/foo', '/bar(.*)'] or `[/^\/foo\/.*$/]`
+ * Path patterns are supported, for example: `['/foo', '/bar{/*extra}'].
+ * 
+ * Note: regular expressions are not supported in path patterns.
+ * 
  * For more information, see: https://github.com/pillarjs/path-to-regexp
  */
 export const createRouteMatcher = (routes: RouteMatcherParam) => {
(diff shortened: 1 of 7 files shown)


dowski:
This is a breaking change for any clients that are using regexes in their route patterns.

sshader (approved):
Nice! Breaking change seems fine here provided we include it in our CHANGELOG.md + update all examples.

I did a quick search for `createRouteMatcher` -- I think there's a code snippet in docs (`docs/pages/authz/nextjs.mdx`) that needs updating?

src/nextjs/server/routeMatcher.ts
```diff
@@ -52,7 +47,10 @@ export type RouteMatcherParam =
...
  *
  * You can use glob patterns to match multiple routes or a function to match against the request object.
- * Path patterns and regular expressions are supported, for example: `['/foo', '/bar(.*)'] or `[/^\/foo\/.*$/]`
+ * Path patterns are supported, for example: `['/foo', '/bar{/*extra}']`.
+ *
+ * Note: regular expressions are not supported in path patterns.
+ *
  * For more information, see: https://github.com/pillarjs/path-to-regexp
  */
 export const createRouteMatcher = (routes: RouteMatcherParam) => {
```
sshader, line 56: You can write type level tests for stuff like this.https://github.com/get-convex/convex-js/blob/main/src/server/schema.test.ts is a fancier example, but seems like it would be nice to test
```
createRouteMatcher("/foo.*")
```
doesn't compile if that's the intent.

One way of doing this is adding a `// @ts-expect-error` comment above the line that's meant to fail typechecking (compilation will succeed if the next line has a type error, and fail otherwise).

dowski:
I'm going to close this out - we'll track any changes in the future in #107.

PR #102, opened 2024-10-17
Add configuration for cookie expiration for Next.js

sshader:
In the current implementation, the cookies we use for Next.js have no expiration which means they're deleted at the end of the browser session (for Chrome, this seems like whenever you quit your browser). A use case where this was particularly disruptive was with mobile web, where closing the app (or maybe even just backgrounding it for long enough) would clear out the cookies.

This PR makes it configurable via an option in `convexAuthNextjsMiddleware`.

In an ideal world, all the configurations would live in one spot (like `auth.ts`), but this is hard in practice since some of the configuration is deployed to Convex while some is deployed to Next.js. It could also be nice to be able to configure this based on a parameter in the sign in request (like the "remember me" check box), but offering this basic configuration is a good start.

### DIFF
--- a/src/nextjs/server/cookies.ts
+++ b/src/nextjs/server/cookies.ts
@@ -2,20 +2,28 @@ import { cookies, headers } from "next/headers";
 import { NextRequest, NextResponse } from "next/server";
 
 export function getRequestCookies() {
-  return getCookieStore(headers(), cookies());
+  return getCookieStore(headers(), cookies(), { maxAge: null });
 }
 
 export function getRequestCookiesInMiddleware(request: NextRequest) {
-  return getCookieStore(headers(), request.cookies);
+  return getCookieStore(headers(), request.cookies, { maxAge: null });
 }
 
-export function getResponseCookies(response: NextResponse) {
-  return getCookieStore(headers(), response.cookies);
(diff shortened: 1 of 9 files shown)


src/nextjs/server/cookies.ts
```diff
@@ -32,6 +47,9 @@ function getCookieStore(
 function getCookieStore(
   requestHeaders: ReturnType<typeof headers>,
   responseCookies: NextResponse["cookies"] | NextRequest["cookies"],
+  cookieConfig: {
+    maxAge: number | null;
+  },
```
dowski, line 52: Are negative numbers valid values for `maxAge`? If not, I suggest throwing an error if one is provided.

dowski: Ah, it looks like zero or negative means expire the cookie.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#max-agenumber

dowski (approved):

src/nextjs/server/cookies.ts
sshader: I'll still add validation, because I think setting all your cookies to immediately expire is not a valid configuration

Issue #92, opened 2024-10-05
Not auto redirect after execute signOut function


tungnc3142:
My project use:

```
"@auth/core": "^0.35.3",
    "@convex-dev/auth": "^0.0.71",
    "@radix-ui/react-icons": "^1.3.0",
    "@radix-ui/react-separator": "^1.1.0",
    "@radix-ui/react-slot": "^1.1.0",
    "class-variance-authority": "^0.7.0",
    "clsx": "^2.1.1",
    "convex": "^1.16.3",
    "lucide-react": "^0.446.0",
    "next": "14.2.13",
    "react": "^18",
    "react-dom": "^18",
    "react-icons": "^5.3.0",
    "tailwind-merge": "^2.5.2",
    "tailwindcss-animate": "^1.0.7"
```

and this is my middleware

```
import {
  convexAuthNextjsMiddleware,
  createRouteMatcher,
  nextjsMiddlewareRedirect
} from '@convex-dev/auth/nextjs/server';

const isPublicPage = createRouteMatcher(['/auth']);

export default convexAuthNextjsMiddleware((request, { convexAuth }) => {
  if (!isPublicPage(request) && !convexAuth.isAuthenticated()) {
    return nextjsMiddlewareRedirect(request, '/auth');
  }
  if (isPublicPage(request) && convexAuth.isAuthenticated()) {
    return nextjsMiddlewareRedirect(request, '/');
  }
});

export const config = {
  // The following matcher runs middleware on all routes
  // except static assets.
  matcher: ['/((?!.*\\..*|_next).*)', '/', '/(api|trpc)(.*)']
};
```

When I execute signOut function from `useAuthActions` hooks the page not auto redirect to auth page but when i use signIn function they still auto redirect to home page
>

kovigv:
Same problem

kovigv:
Looks like you need to add some extra code. For example:

```typescript
import { useRouter } from "next/navigation";

const router = useRouter()

  const onPasswordSignIn = (e: React.FormEvent<HTMLFormElement>) => {
    e.preventDefault()
    setPending(true)
    signIn("password", {email, password, flow: "signIn"})
      .catch(() => {
        setError("Invalid email or password")
      })
      .finally(() => {
        setPending(false)
        router.push("/")
      })
  }
```

So router.push("/") will be a fix for that
You should add it on 
onPasswordSignIn
onPasswordSignUp
and on signOut
```typescript
"use client";

[...]

PR #104, opened 2024-10-20
Set up Playwright tests for Nextjs

xixixao:
- Adds a "secret" auth method in for e2e tests (should be ok, I think this example is only used for testing here)
- Fixes the isLocalhost check to work in Playwright in all browsers (this took me forever to figure out, it worked fine in Firefox...)
- Doesn't add the tests to CI or the main package.json, that should be easy to do though
- Adds test for sign-out and a failed sign-in - unfortunately the setup in `test-nextjs` doesn't reproduce https://github.com/get-convex/convex-auth/issues/92, which is what I was trying to debug
- I decided it's better to keep the "secret" login UI on the same route as the normal one, so that the middleware doesn't have to change and we're testing its behavior

----


dowski (approved):
Thanks for this, Michal!

PR #108, opened 2024-10-23
Handle password validation failure in the custom sign-up example

dowski:
Uses a shared constant value between the backend auth config and the frontend and sends it in a ConvexError when password validation fails.

----


test/src/auth/SignInWithPassword.tsx
```diff
@@ -20,6 +22,8 @@ export function SignInWithPassword({
   const { signIn } = useAuthActions();
   const [flow, setFlow] = useState<"signIn" | "signUp">("signIn");
   const { toast } = useToast();
+  const [passwordRequirementsColor, setPasswordRequirementsColor] =
+    useState("text-gray-500");
```
xixixao, line 26: We should not encourage this pattern (this code is linked from docs).

Instead use something like `const [error, setError]` and keep the mapping to colors in the JSX.

UX-wise though it might be better to show a different toast, because the color won't change on subsequent submits. This will avoid the need for more state altogether.

test/src/auth/SignInWithPassword.tsx
```diff
@@ -34,11 +38,18 @@ export function SignInWithPassword({
...
           .catch((error) => {
             console.error(error);
-            const title =
-              flow === "signIn"
-                ? "Could not sign in, did you mean to sign up?"
-                : "Could not sign up, did you mean to sign in?";
-            toast({ title, variant: "destructive" });
+            if (
+              error instanceof ConvexError &&
+              error.data === passwordValidationFailed
```
xixixao, line 43:
```suggestion
              error.data === INVALID_PASSWORD
```

test/convex/errors.ts
```diff
@@ -0,0 +1 @@
+export const passwordValidationFailed = "password_validation_failed";
```
xixixao, line 1:
```suggestion
export const INVALID_PASSWORD = "INVALID_PASSWORD";
```

test/src/auth/SignInWithPassword.tsx
dowski: Sounds good, switched to using the toast for simplicity.

test/convex/errors.ts
dowski: Thanks for the suggestion!

sshader (approved):

PR #109, opened 2024-10-23
Update the lock file to point to a common @types/react version

dowski:
Previously the root package and the test-nextjs package pointed at different versions and I think that was causing the following error:

```
% npm run build

> template-nextjs@0.1.0 build
> next build

  ▲ Next.js 14.2.5
  - Environments: .env.local

   Creating an optimized production build ...
 ⚠ Found lockfile missing swc dependencies, patching...
 ⚠ Found lockfile missing swc dependencies, patching...
 ⚠ Found lockfile missing swc dependencies, patching...
 ⚠ Lockfile was successfully patched, please run "npm install" to ensure @next/swc dependencies are downloaded
 ✓ Compiled successfully
   Linting and checking validity of types  ..Failed to compile.

./components/ui/toggle-group.tsx:28:7
Type error: Type 'React.ReactNode' is not assignable to type 'import("/Users/christian/Development/Convex/convex-auth/test-nextjs/node_modules/@types/react/index").ReactNode'.
  Type 'bigint' is not assignable to type 'ReactNode'.

  26 |   >
  27 |     <ToggleGroupContext.Provider value={{ variant, size }}>
> 28 |       {children}
     |       ^
  29 |     </ToggleGroupContext.Provider>
  30 |   </ToggleGroupPrimitive.Root>
  31 | ));
```

----

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1774,10 +1774,11 @@
       "dev": true
     },
     "node_modules/@types/react": {
-      "version": "18.3.3",
-      "resolved": "https://registry.npmjs.org/@types/react/-/react-18.3.3.tgz",
-      "integrity": "sha512-hti/R0pS0q1/xx+TsI73XIqk26eBsISZ2R0wUijXIngRK9R/e7Xw/cXVxQK7R5JjW+SV4zGcn5hXjudkN/pLIw==",
+      "version": "18.3.12",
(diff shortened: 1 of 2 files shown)


sshader (approved):
Can we bump the required version in the `package.json` dev dependencies instead? I think that's a slightly better source of truth than the lock file? (but also this is fine if it builds)

dowski:
Yeah, good idea to have it specified directly in the `package.json`.

Issue #113, opened 2024-10-25
NextJS 15 support


mez:
Just tried to update to Next15, the Async Request APIs breaking changes did break `convexAuth` specifically cookie and token requesting area.

```js
[Error: Route "/" used `headers().get('Host')`. `headers()` should be awaited before using its value. Learn more: https://nextjs.org/docs/messages/sync-dynamic-apis]
[Error: Route "/" used `cookies().get('__convexAuthJWT')`. `cookies()` should be awaited before using its value. Learn more: https://nextjs.org/docs/messages/sync-dynamic-apis]
[Error: Route "/" used `cookies().get('__convexAuthRefreshToken')`. `cookies()` should be awaited before using its value. Learn more: https://nextjs.org/docs/messages/sync-dynamic-apis]
[Error: Route "/" used `headers().get('Host')`. `headers()` should be awaited before using its value. Learn more: https://nextjs.org/docs/messages/sync-dynamic-apis]
[Error: Route "/" used `cookies().get('__convexAuthJWT')`. `cookies()` should be awaited before using its value. Learn more: https://nextjs.org/docs/messages/sync-dynamic-apis]
```

To reproduce, just update to Next 15 while using convexAuth.

thomasballinger:
sounds like https://nextjs.org/blog/next-15#async-request-apis-breaking-change

AndrewDonelson:
wow :(
ETA on this? one of my project was giving me issues and i upgraded node, npm and next 14->15 and now none of my projects work. I really do not want to go back to the previous versions and ALL of my apps use convex + convex auth

thomasballinger:
Here's a Next.js 15 demo using a branch of this library (published as @convex-dev/auth@0.0.75-alpha.1) and a new convex alpha (convex@1.17.1-alpha.1): https://github.com/get-convex/convex-auth-test-next15

These are experimental builds, if you try them please let us know how they are!

Before merging https://github.com/get-convex/convex-auth/pull/118 I'd like to get some feedback.

JohnKesko:
@thomasballinger 
Just tried it out, here is some feedback.
(Note I am also using @auth/core": "^0.37.4)

So this installs fine while also having @auth/core:
"@convex-dev/auth": "0.0.75-alpha.1"

But updating "convex": "^1.17.0" to  "convex": "1.17.1-alpha.1" does not work in same context with @auth/core.
[...]

PR #118, opened 2024-11-10
Next15 support

cmaycumber:
Adds support for breaking changing in NextJS 15 fixing #113 

----

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+- BREAKING: `convexAuthNextjsToken()` and `isAuthenticatedNextjs()` now return
+  promises so must be `await`ed.
+- Support for Next.js 15.
+- Update convex peer dependency to ^1.17.0
+
 ## 0.0.74
 
 - Fix to header propagation in Next.js middleware
--- a/docs/pages/authz/nextjs.mdx
+++ b/docs/pages/authz/nextjs.mdx
@@ -1,11 +1,9 @@
 import { Callout } from "nextra/components";
 
-# Server-side authentication in Next.js 14
+# Server-side authentication in Next.js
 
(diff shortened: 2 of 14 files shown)


thomasballinger:
Hm it's a bummer for this change to break everyone still using Next.js 14. Maybe we could change some APIs to allow promises, like in the docs change here allowing a `Promise<string>` for the token. Otherwise we're requiring a change on upgrade.

thomasballinger:
Let's go ahead and require that Next.js 14 users have to make changes, but let's change the TypeScript types here to make sure we don't call `.then()` on something that is only a promise in Next.js 15.

cmaycumber:
Got it. This was a bit quick and dirty. I'll try to clean up some of the awaits as well to make the code a little more readable.

thomasballinger:
i can pick it up, feel free to leave as is if you haven't! But that sounds great

thomasballinger:
OK there's more to do here, but this is great so far.
- create an example repo (or update test-nextjs in this one) to show this working.
- open up the peerDependencies like
```
  "peerDependencies": {
    "next": "^14.0.3 || ^15.0.0",
    "react": "^18 || ^19.0.0-0",
  },
  ```

But let's not block an alpha on all that. @cmaycumber is this working for you in a Next.js 15 app? Do you see peer dependency warnings you have to ignore?

thomasballinger:
I'm going to set up a Next.js 15 example, get it running automated tests, point it at this PR, and if everything works with this this PR seems good to publish as an alpha.

thomasballinger:
Here's a Next.js 15 demo using this branch (published as `@convex-dev/auth@0.0.75-alpha.1`) and a new convex alpha (`convex@1.17.1-alpha.1`): https://github.com/get-convex/convex-auth-test-next15

thomasballinger:
Thanks @cmaycumber, merged in as https://github.com/get-convex/convex-auth/pull/125

PR #119, opened 2024-11-12
Upgrade to @auth/core 0.37.3

dowski:
This is a minimal approach to switching to `@auth/core` 0.37+. It pulls in a bit more code, but lets us run using the new 3.x version of `oauth4webapi` as well.

The test suites (`test/` and `test-nextjs/`) succeed. I plan on trying out a few providers manually with this change, but thought I'd put the code up for review in the meantime.

----

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "server-only": "^0.0.1"
   },
   "peerDependencies": {
-    "@auth/core": "^0.36.0",
+    "@auth/core": "^0.37.0",
     "convex": "^1.14.4",
     "react": "^18.2.0"
   },
--- a/src/server/oauth.ts
+++ b/src/server/oauth.ts
@@ -260,7 +260,7 @@ export async function handleOAuthCallback(
       codeGrantResponse;
   }
 
-  let profile: unknown = {};
+  let profile: any = {};
   let tokens: TokenSet & Pick<Account, "expires_at">;
 
   let nonce;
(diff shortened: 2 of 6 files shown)


src/server/oauth.ts
```diff
@@ -221,10 +260,6 @@ export async function handleOAuthCallback(
       codeGrantResponse;
   }
 
-  if (o.parseWwwAuthenticateChallenges(codeGrantResponse)) {
```
sshader, line 224: Does this move somewhere else?

src/server/oauth.ts
```diff
@@ -236,30 +271,24 @@ export async function handleOAuthCallback(
...
     }
-    const result = await o.processAuthorizationCodeOpenIDResponse(
+    const result = await o.processAuthorizationCodeResponse(
       server,
       client,
       codeGrantResponse,
-      nonce ?? o.expectNoNonce,
+      {
+        expectedNonce: nonce ?? o.expectNoNonce,
+        requireIdToken: true,
```
sshader, line 280: Do we understand what this does / why it's safe to change here?

package.json
```diff
@@ -71,7 +71,7 @@
     "server-only": "^0.0.1"
   },
   "peerDependencies": {
-    "@auth/core": "^0.36.0",
+    "@auth/core": "^0.37.0",
```
sshader, line 74: If we land this, we should also update the docs that tell you to install `@auth/core@0.36.0`

sshader (approved):
I'm down to try it (with a bunch of manual testing). It does make me nervous to patch in changes `@auth/core` that we perhaps don't fully understand

src/server/oauth.ts
dowski: `oauth4webapi` 3.0.0 changed the API and replaced a couple functions with this one.

> processAuthorizationCodeOpenIDResponse() method was removed in favour of processAuthorizationCodeResponse()
> processAuthorizationCodeOAuth2Response() method was removed in favour of processAuthorizationCodeResponse()

https://github.com/panva/oauth4webapi/releases/tag/v3.0.0

These extra args passed here are for OIDC - Auth.js does it in a similar way:

https://github.com/nextauthjs/next-auth/blob/1c9bcdd0c4538a852f8d2b2b7c60eb962e3a50eb/packages/core/src/lib/actions/callback/oauth/callback.ts#L215-L223

package.json
dowski: I'm guessing you mean `0.37.0`.

dowski (commented):
So far I've manually tested this with Github, Google and Apple (the latter required a few tweaks to work, I think that was required whether we upgrade to `@auth/core` 0.37.0 or not).

package.json
sshader: We should update the docs that tell you to install `@auth/core@0.36.0` to tell you to install `@auth/core@0.37.0`

dowski: Yep! Updated!

PR #125, opened 2024-11-20
Next15 support

thomasballinger:


thomasballinger:
Iterating separately from Chadwick's branch, https://github.com/get-convex/convex-auth/pull/118

sshader (approved):
So we've gotten some testing with Next 15 and the alpha version with this code -- as long as we've also manually tested that Next 14 still works (after adding all the `await`s) I'm down to ship it

thomasballinger:
Just manually tested, looks good with a Next.js 14 site.

docs/pages/authz/nextjs.mdx
```diff
@@ -28,10 +26,10 @@ const isSignInPage = createRouteMatcher(["/signin"]);
 const isProtectedRoute = createRouteMatcher(["/product(.*)"]);
 
 export default convexAuthNextjsMiddleware((request, { convexAuth }) => {
```
edproton, line 28: Is lacking `async` here.

edproton (commented):
Improve docs

PR #129, opened 2024-12-03
Use a Convex function for Next.js auth checks

dowski:
If the auth token doesn't validate on the Convex
backend, then it's considered invalid and thus
unauthenticated.

The old check just validated that there was some token
cookie but not that it was valid for use.

----

### DIFF
--- a/src/nextjs/server/index.tsx
+++ b/src/nextjs/server/index.tsx
@@ -233,7 +233,11 @@ export function convexAuthNextjsMiddleware(
             isAuthenticated: async () => {
               const cookies = await getRequestCookiesInMiddleware(request);
               try {
-              return await fetchQuery("auth:isAuthenticated" as any as IsAuthenticatedQuery, {}, {token: cookies.token ?? undefined});
+                return await fetchQuery(
+                  "auth:isAuthenticated" as any as IsAuthenticatedQuery,
+                  {},
+                  { token: cookies.token ?? undefined },
+                );
               } catch {
                 return false;
               }
--- a/src/nextjs/server/index.tsx
+++ b/src/nextjs/server/index.tsx
@@ -1,5 +1,6 @@
 import "server-only";
 
+import { fetchQuery } from "convex/nextjs";
 import { NextMiddlewareResult } from "next/dist/server/web/types";
 import {
   NextFetchEvent,
@@ -20,6 +21,7 @@ import {
   setAuthCookies,
   setAuthCookiesInMiddleware,
 } from "./utils.js";
+import { IsAuthenticatedQuery } from "../../server/implementation/index.js";
 
 /**
  * Wrap your app with this provider in your root `layout.tsx`.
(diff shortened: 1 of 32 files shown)


test-nextjs/convex/auth.ts
```diff
@@ -3,7 +3,7 @@ import GitHub from "@auth/core/providers/github";
 import { convexAuth } from "@convex-dev/auth/server";
 import { internal } from "./_generated/api";
 
-export const { auth, signIn, signOut, store } = convexAuth({
+export const { auth, signIn, signOut, store, isAuthenticated } = convexAuth({
```
dowski, line 6: I think one thing I don't like about this is that it expands the required members of the `auth:` namespace.

Maybe instead I can bundle it in the `auth` object that's already unpacked/exported here?

dowski: That doesn't seem to work.

src/nextjs/server/index.tsx
```diff
@@ -111,7 +113,7 @@ export function isAuthenticatedNextjs() {
  */
 export type ConvexAuthNextjsMiddlewareContext = {
   getToken: () => string | undefined;
-  isAuthenticated: () => boolean;
+  isAuthenticated: () => Promise<boolean>;
```
sshader, line 116: You probably need a rebase since I think these are already async https://github.com/get-convex/convex-auth/pull/125

sshader (commented):
(oops forgot to click submit yesterday)

src/nextjs/server/index.tsx
dowski: 👍

src/nextjs/server/index.tsx
```diff
@@ -227,9 +229,13 @@ export function convexAuthNextjsMiddleware(
               const cookies = getRequestCookiesInMiddleware(request);
               return cookies.token ?? undefined;
             },
-            isAuthenticated: () => {
+            isAuthenticated: async () => {
               const cookies = getRequestCookiesInMiddleware(request);
-              return (cookies.token ?? undefined) !== undefined;
+              try {
+              return await fetchQuery("auth:isAuthenticated" as any as IsAuthenticatedQuery, {}, {token: cookies.token ?? undefined});
```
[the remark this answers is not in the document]
dowski: 👍

test-nextjs/e2e-tests/route-handler-with-auth.spec.ts
```diff
@@ -0,0 +1,33 @@
+import { test, expect, Page } from "@playwright/test";
```
thomasballinger, line 1: awesome

test-nextjs/playwright.config.ts
```diff
@@ -77,5 +77,7 @@ export default defineConfig({
       : "npm run dev:frontend",
     url: "http://127.0.0.1:3000",
     reuseExistingServer: !process.env.CI,
+    stdout: "ignore", // Set to "pipe" if you're debugging a failing test.
```
thomasballinger, line 80: is it too loud? if this cost is just scrolling past logs seems useful to just have it?

thomasballinger (approved):

test-nextjs/playwright.config.ts
dowski: It adds about 100 lines of log output to a successful test run. But I guess others working on this might not even know to come in here and tweak this to debug a failing test - you're probably right that it's better to have on by default.

dowski:
Thanks for the reviews!

PR #130, opened 2024-12-04
Pull in Auth.js code that we plan to fork

dowski:
The code is from commit https://github.com/nextauthjs/next-auth/blob/5af1f30a32e64591abc50ae4d2dba4682e525431. Some of the files were renamed and don't match the original path in the next-auth repo. That will be clarified in comments when our changes are applied to the fork.

----


dowski:
Since this was a part of #121 that was approved and just split out for the clear repo history entry, I'm going to merge it.

PR #138, opened 2024-12-12
downgrade userId

ianmacartney:
Just exploring what it'd look like if userId was a string with a few annotations on the edges

----

### DIFF
--- a/src/providers/Password.ts
+++ b/src/providers/Password.ts
@@ -137,7 +137,6 @@ export function Password<DataModel extends GenericDataModel>(
       const { email } = profile;
       const secret = params.password as string;
       let account: GenericDoc<DataModel, "authAccounts">;
-      let user: GenericDoc<DataModel, "users">;
       if (flow === "signUp") {
         if (secret === undefined) {
           throw new Error("Missing `password` param for `signUp` flow");
@@ -149,7 +148,7 @@ export function Password<DataModel extends GenericDataModel>(
           shouldLinkViaEmail: config.verify !== undefined,
           shouldLinkViaPhone: false,
         });
-        ({ account, user } = created);
+        ({ account } = created);
       } else if (flow === "signIn") {
         if (secret === undefined) {
           throw new Error("Missing `password` param for `signIn` flow");
@@ -161,7 +160,7 @@ export function Password<DataModel extends GenericDataModel>(
         if (retrieved === null) {
           throw new Error("Invalid credentials");
         }
-        ({ account, user } = retrieved);
+        ({ account } = retrieved);
         // START: Optional, support password reset
       } else if (flow === "reset") {
         if (!config.reset) {
@@ -226,7 +225,7 @@ export function Password<DataModel extends GenericDataModel>(
         });
       }
       // END
-      return { userId: user._id };
+      return { userId: account.userId as string };
     },
     crypto: {
       async hashSecret(password: string) {
--- a/src/server/implementation/mutations/retrieveAccountWithCredentials.ts
+++ b/src/server/implementation/mutations/retrieveAccountWithCredentials.ts
@@ -17,7 +17,7 @@ type ReturnType =
   | "InvalidAccountId"
   | "TooManyFailedAttempts"
   | "InvalidSecret"
-  | { account: Doc<"authAccounts">; user: Doc<"users"> };
+  | { account: Doc<"authAccounts"> };
 
 export async function retrieveAccountWithCredentialsImpl(
   ctx: MutationCtx,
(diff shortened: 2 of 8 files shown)


src/providers/Anonymous.ts
```diff
@@ -28,7 +28,13 @@ import { Value } from "convex/values";
 /**
  * The available options to an {@link Anonymous} provider for Convex Auth.
  */
-export interface AnonymousConfig<DataModel extends GenericDataModel> {
+export interface AnonymousConfig<
+  DataModel extends GenericDataModel,
+  UserDocument extends Record<string, Value> = DocumentByName<
+    DataModel,
+    "users"
+  >,
```
ianmacartney, line 36: I wonder if we should just have this be the same as other providers, vs. returning a specific type.. not sure if this is causing the type issues

src/providers/Anonymous.ts
```diff
@@ -66,13 +72,13 @@ export function Anonymous<DataModel extends GenericDataModel>(
     id: "anonymous",
     authorize: async (params, ctx) => {
       const profile = config.profile?.(params, ctx) ?? { isAnonymous: true };
-      const { user } = await createAccount(ctx, {
+      const { account } = await createAccount(ctx, {
```
ianmacartney, line 75: one concern is that folks who have made their own providers may be tripped up by `createAccount` not returning the user anymore. This would be a breaking change technically

src/providers/Password.ts
```diff
@@ -48,7 +49,10 @@ import { Scrypt } from "lucia";
 /**
  * The available options to a {@link Password} provider for Convex Auth.
  */
-export interface PasswordConfig<DataModel extends GenericDataModel> {
+export interface PasswordConfig<
+  DataModel extends GenericDataModel,
+  UsersTable extends TableNamesInDataModel<DataModel> = "users",
```
ianmacartney, line 54: could just do the `UsersDoc` like anonymous to be more flexible, but not sure which is more convenient / types better

src/server/implementation/mutations/retrieveAccountWithCredentials.ts
```diff
@@ -60,8 +60,6 @@ export async function retrieveAccountWithCredentialsImpl(
   }
   return {
     account: existingAccount,
-    // TODO: Ian removed this
-    user: (await ctx.db.get(existingAccount.userId))!,
```
ianmacartney, line 64: I don't remember why this was added back. I thought it wasn't needed beyond getting the userId, but there's a decent amount of untyped code..

src/server/implementation/users.ts
```diff
@@ -54,13 +54,23 @@ async function defaultCreateOrUpdateUser(
...
       existingUserId,
       ...args,
     });
+  } else {
+    if (
+      existingUserId &&
+      ctx.db.normalizeId("users", existingUserId) === null
+    ) {
+      throw new Error(
+        `User ID \`${existingUserId}\` is not in the \`users\` table`,
```
ianmacartney, line 71: Could use a better message about defining your own `createOrUpdateUser` to use another table

src/server/types.ts
```diff
@@ -341,7 +341,9 @@ export type PhoneUserConfig<
 /**
  * Similar to Auth.js Credentials config.
  */
-export type ConvexCredentialsConfig = ConvexCredentialsUserConfig<any> & {
+export type ConvexCredentialsConfig<
+  DataModel extends GenericDataModel = GenericDataModel,
+> = ConvexCredentialsUserConfig<DataModel> & {
```
ianmacartney, line 346: this type might be why there's now circular types issues..?

thomasballinger:
the proposal here (based on this PR and chatting in person) is
- ConvexAuthConfig is parameterized by a string type, `Id<"users">` by default. Any use of `GenericId<"users">` is now this type, which doesn't even need to be an Id or refer to a table at all.
- Functions like `upsertUserAndAccount` and `invalidateSessions` that accepted `GenericId<"users">` now accept a string
- We want functions like `getAuthUserId` that returned `GenericId<"users">` to return a string, but that's not backward-compatible, so we add an optional type parameter that is still `GenericId<"users">` by default.

This makes it possible to change the name of the "users" table by
- passing a different ID type to ConvexAuthConfig
- wrapping `getAuthUserId` in a helper that casts it to your new ID type
- passing a createOrUpdateUser callback which returns a GenericId<"myOtherUsersTable">

This also makes the advanced use case of completely customizing how users are stored possible (maybe they're in two tables, it's they're in an external system, etc.) by
- passing a string type instead of an ID type to ConvexAuthConfig
- passing a createOrUpdateUser callback which returns a string instead of a GenericId<"users">

As a long-term goal, if all interaction with the users table is developer-mediated then there are no Convex Auth-imposed constraints on the users table. Assumptions about the users table (described [here](https://labs.convex.dev/auth/setup/schema#customizing-the-users-table)) wouldn't be needed.

thomasballinger:
We'll also want to make writing a createOrUpdateUser callback simpler if we do this.

PR #141, opened 2024-12-14
fix module resolution error for checks import

erquhart:
A module resolution error occurs due to no extension on an import from convex-test. Not certain why this is required, but adding an extension to align with the other imports fixes the issue.

Reference discord post: https://discord.com/channels/1019350475847499849/1317446619775373393

----


thomasballinger:
Cool, later let's add a linter for this too.

Issue #53, opened 2024-08-14
Support custom redirect domains


ianmacartney:
So you don't have a .convex.site url in the OAuth flow.

It doesn't necessarily need to replace issuer & other uses, but being able to redirect to it would be nice

jabra:
I hope i'm adding to the right issue as suggested by @thomasballinger .
I setup my app with convex auth and user authenticating using google provider.
When click on the google login button, I see the convex.site on the google auth page 
![image](https://github.com/user-attachments/assets/5df73018-7973-4a4e-abf7-792efd294871)


I tried the custom domain route as described in this article https://docs.convex.dev/production/hosting/custom
![image](https://github.com/user-attachments/assets/aa1462b6-3242-475f-a62f-451c5c4a8413)

but I cannot override the CONVEX_SITE_URL 
![image](https://github.com/user-attachments/assets/8895fdc2-4277-43e8-991b-6596771b3913)

I'm not even sure this is the right approach to get google auth screen to not display the *.convex.site as origin but my own site.

dowski:
Hey, @jabra - just chiming in here with an idea. 

If you verify your Google OAuth app, it should show the app name and logo that you defined in your app's branding settings instead of the domain.

dowski:
Here's their docs on [verification](https://support.google.com/cloud/answer/13463073?hl=en).

dowski:
Docs for this are here: https://labs.convex.dev/auth/advanced#custom-callback-and-sign-in-urls

We'll need to cut a new release before it's available for general use.

PR #142, opened 2024-12-17
A couple tweaks to allow a CUSTOM_AUTH_SITE_URL

dowski:
This will allow a custom domain for signin and callback URLs. OAuth providers like Google (which show the domain) will use this instead of showing `happy-animal-123.convex.site` in the OAuth consent screen.

This should help to resolve https://github.com/get-convex/convex-auth/issues/53.

----


thomasballinger (approved):
That's slick! If there's any confusing or possible mismatch issue we can anticipate interested in addressing it next, this PR lgtm though

PR #144, opened 2024-12-18
Make isAuthenticated in middleware fail more loudly.

thomasballinger:
Make the Next.js middleware function `isAuthenticated` fail more loudly. Previously it returned false in the case of a Convex backend that didn't expose an endpoint called `auth:isAuthenticated`, now it throws an error. This should help people with the migration required for 0.0.76.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+- Next.js middleware function `isAuthenticated` fails more loudly; previously it
+  returned false in the case of a Convex backend that didn't expose an endpoint
+  called `auth:isAuthenticated`, now it throws an error. This should help people
+  with the migration required for 0.0.76.
+
 ## 0.0.78
 
 - Add support for
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,6 @@
   called `auth:isAuthenticated`, now it throws an error. This should help people
   with the migration required for 0.0.76.
 
-- Export `SignInAction` and `SignOutAction` types.
-
 ## 0.0.77
 
 - Fix syntax of an import to work with convex-test.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+- Next.js middleware function `isAuthenticated` fails more loudly; previously it
+  returned false in the case of a Convex backend that didn't expose an endpoint
+  called `auth:isAuthenticated`, now it throws an error. This should help people
+  with the migration required for 0.0.76.
+
 ## 0.0.78
```
dowski, line 10: Technically this is unreleased too, I think - I just speculatively added the next version # when I updated this file earlier.

dowski (approved):
LG, but see other comments.

src/nextjs/server/index.tsx
```diff
@@ -308,7 +308,12 @@ async function isAuthenticated(token: string | null): Promise<boolean> {
...
     );
-  } catch {
-    return false;
+  } catch (e: any) {
+    if (e.message.includes("Could not find public function")) {
+      throw new Error(
+        "Server Error: could not find api.auth.isAuthenticated. convex-auth 0.0.76 introduced a new export in convex/auth.ts. Add `isAuthenticated` to the list of functions returned from convexAuth(). See convex-auth changelog for more https://github.com/get-convex/convex-auth/blob/main/CHANGELOG.md",
+      );
+    }
+    throw e;
```
dowski, line 317: I think you need to return false here. Otherwise errors like sending an invalid JWT to the Convex backend will also throw (which is why that one test is failing, I think).

CHANGELOG.md
thomasballinger: ah gotcha

src/nextjs/server/index.tsx
thomasballinger: Ah for the token expired and invalid token cases, got it. Maybe we can pick those out too, for now I'll log whatever error it is.

PR #163, opened 2025-02-10
retry fetch token on network error

erquhart:
A single token fetch failing due to network error can lead to an erroneous unauthenticated client state (where refresh/reload leads to authenticated state, without signing in). One example of this is a react native app fetching a token, and the app is moved to background before the request resolves. At least on iOS, when the app is active again, the fetch will fail due to network error and the user will see an unauthenticated state.

This PR:
- splits out the `verifyCode` portion of `verifyCodeAndSetToken`
- allows up to two retries if fetch fails due to network error


src/react/client.tsx
```diff
@@ -164,16 +167,42 @@ export function AuthProvider({
...
   }, [setToken]);
 
+  const verifyCode = useCallback(
+    async (
+      args: { code: string; verifier?: string } | { refreshToken: string },
+    ) => {
+      let lastError;
+      // Retry the call if it fails due to a network error.
+      let retries = FETCH_TOKEN_RETRIES;
+      while (retries > 0) {
```
ianmacartney, line 177: Should we add some backoff here? e.g. https://github.com/get-convex/llama-farm-chat/blob/e81130e439e3797b65dc05e3a3473c258744bec8/shared/llm.ts#L311

erquhart: Yeah, will add

src/react/client.tsx
```diff
@@ -18,6 +18,11 @@ import type {
   ConvexAuthActionsContext as ConvexAuthActionsContextType,
   TokenStorage,
 } from "./index.js";
+import { isNetworkError } from "./is_network_error.js";
+
+// Retry after this much time, based on the retry number.
+const RETRY_BACKOFF = [100, 1000]; // In ms
+const RETRY_JITTER = 10; // In ms
```
sshader, line 25: This is not a lot of jitter. In other spots we do jitter proportional to the backoff -- checkout [this](https://github.com/get-convex/convex-js/blob/26f3416b69095eba4a87df0f09184920a33393cb/src/browser/sync/web_socket_manager.ts#L505)

src/react/client.tsx
```diff
@@ -164,16 +169,44 @@ export function AuthProvider({
     return () => browserRemoveEventListener("storage", listener);
   }, [setToken]);
 
+  const verifyCode = useCallback(
+    async (
+      args: { code: string; verifier?: string } | { refreshToken: string },
+    ) => {
+      let lastError;
+      // Retry the call if it fails due to a network error.
```
sshader, line 177:
```suggestion
      // Retry the call if it fails due to a network error.
      // This is especially common in mobile apps where an app is backgrounded while making a call and hits a network error, but will succeed with a retry once the app is brought to the foreground.
```
(wordsmith however you like + correct anything I got wrong here)

src/react/is_network_error.ts
```diff
@@ -0,0 +1,60 @@
+/**
+ * is-network-error@1.1.0
```
sshader, line 2: just curious -- why copying this in vs. adding as an npm dependency? (I think the latter is slightly easier to keep up to date automatically?)

sshader (approved):

src/react/is_network_error.ts
erquhart: Can definitely add as a dependency, my only hesitations were whether adding a dependency would be welcome, and secondarily that I've seen Sindre's pure ESM packages cause problems with projects. Don't think that'll be an issue here, though - will update.

src/react/client.tsx
erquhart: Gah, thought I pushed an update to this - will fix.

erquhart:
@sshader thanks for the comments, addressed those. I also included something I forgot to push earlier that was used in troubleshooting this issue - defaulting to use the Convex Client logger if no logger is explicitly provided. Let me know if this should be dropped, I can open a separate PR for it or just leave it out.

src/react/client.tsx
```diff
@@ -164,16 +170,47 @@ export function AuthProvider({
...
+            "code" in args
+              ? { params: { code: args.code }, verifier: args.verifier }
+              : args,
+          );
+        } catch (e) {
+          lastError = e;
+          if (!isNetworkError(e)) {
+            break;
+          }
+          const wait = RETRY_BACKOFF[retry] + RETRY_JITTER * Math.random();
```
sshader, line 196: Can we explicitly check that `retry < RETRY_BACKOFF.length` and break? I'm actually unclear what `wait` becomes in that case

sshader (approved):
The `logger` addition seems fine (we could probably eventually clean up some of the `verbose` vs. `logger` stuff, but this all seems fine)

src/react/client.tsx
erquhart: There's an explicit check in the while condition a few lines up, but I can refactor if it's unclear

sshader: oh I swear I saw one before and was wondering where it went. I'm just blind, sorry

erquhart: Lol no problem!

erquhart:
(bump)

PR #170, opened 2025-02-27
Added missing import in nextjs docs page

sanidhyy:
Fixed a missing import in the [Next.js docs page](https://labs.convex.dev/auth/authz/nextjs#calling-authenticated-mutations-and-actions) under the 'Calling authenticated mutations and actions' section. Added the missing `convexAuthNextjsToken` import in the code example.

----


sshader:
Thanks for the fix!

PR #183, opened 2025-03-20
Add networkless path for isAuthenticated

sujayakar:
Following Clerk's lead, add an optional environment variable that lets developers configure Next to not require hitting Convex for validating its token.

Verification is a little more permissive than https://github.com/get-convex/convex-backend/blob/main/crates/authentication/src/lib.rs#L188 but I think it's okay to not check that the issuer and audience match. We can add both of these checks via another environment variable if needed.

Question: Should I add this verification to the methods like `convexAuthNextjsToken` that pull out the token from cookies? I think it's fine to only check when we're making a decision (i.e. `isAuthenticated`) but I'm curious what y'all think.

### DIFF
--- a/src/nextjs/server/index.tsx
+++ b/src/nextjs/server/index.tsx
@@ -22,6 +22,7 @@ import {
   setAuthCookiesInMiddleware,
 } from "./utils.js";
 import { IsAuthenticatedQuery } from "../../server/implementation/index.js";
+import { jwtVerify, createLocalJWKSet } from "jose"
 
 /**
  * Wrap your app with this provider in your root `layout.tsx`.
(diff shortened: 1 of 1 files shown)


src/nextjs/server/index.tsx
```diff
@@ -302,6 +303,21 @@ async function isAuthenticated(token: string | null): Promise<boolean> {
   if (!token) {
     return false;
   }
+
+  // First, try using the JWKS from the environment variable to do token
+  // verification locally (networkless mode).
+  try {
+    const envJwks = process.env.CONVEX_AUTH_JWKS
```
sshader, line 310: I was going to ask if this is possible to configure via an argument in the provider or middleware, but I think the answer is "no" since these get wrapped in things that are imported directly.

src/nextjs/server/index.tsx
```diff
@@ -302,6 +303,21 @@ async function isAuthenticated(token: string | null): Promise<boolean> {
...
+    if (envJwks) {
+      const jwkSet = createLocalJWKSet(JSON.parse(envJwks))
+      const verifiedToken = await jwtVerify(token, jwkSet);
```
sshader, line 313: Can we put this + the `JSON.parse` in a `try/catch` (idk if `jwtVerify` can throw, but `JSON.parse` definitely can)

src/nextjs/server/index.tsx
```diff
@@ -302,6 +303,21 @@ async function isAuthenticated(token: string | null): Promise<boolean> {
...
+      return !!verifiedToken.payload.sub;
+    }
+  } catch (error: any) {
+    console.error("Error verifying token", error);
```
sshader, line 317:
```suggestion
    console.error("Error verifying token using process.env.CONVEX_AUTH_JWKS", error);
```
(we could also say something like "make sure this matches the value set in your Convex `JWKS` environment variable")

sshader (approved):
Add docs too? Otherwise seems good.

Re: skipping the check when getting to token -- it's possible that someone does something like `const isAuthenticated = (await convexAuthNextJsToken()) !== null` and is sad. But I think the common case will be to feed the token into a function that will throw / return nothing if the token is invalid.

PR #186, opened 2025-04-01
[cli] Run without prompts if in a non-interactive environment

sshader:
If this is run in a non-interactive terminal, accept the defaults (or crash if there's no safe default) instead of just hanging.


test/convex/auth.ts
```diff
@@ -1,4 +1,4 @@
-import { INVALID_PASSWORD } from "./errors.js"
+import { INVALID_PASSWORD } from "./errors.js";
```
sshader, line 1: drive by lint fix

sshader:
(tests pass locally and are almost certainly unrelated to my change here, so merging)

PR #188, opened 2025-04-02
Accept optional shouldHandleCode config function to filter code param handling

erquhart:
Adds a middleware config function that allows the user to opt out of Convex Auth intercepting the code param.

Fix user validated here: https://discord.com/channels/1019350475847499849/1355618288964735165/1356791107815604235


sshader (approved):

srondss:
@erquhart 
Screenshot 2025-05-08 at 13 28 47

It seems like this was added in `src/react/client.tsx` but what's accessible to the user is actually `src/react/index.tsx` which doesn't have the shouldHandleCode prop. 

I could be wrong but not sure how to access this otherwise.

Issue #103, opened 2024-10-19
Cannot read properties of null (reading 'redirect') when using Password Provider


xiaoqingming18:
I encountered an issue while developing the user registration feature using the **Password Provide**r from Convex Auth in a Next.js project. Upon submitting the registration form, the following error occurs:

`Cannot read properties of null (reading 'redirect')`

![PixPin_2024-10-19_20-19-50](https://github.com/user-attachments/assets/18c3130d-0e93-49d0-a706-b48dd3e0242a)

Additionally, the user registration data is not being submitted to the database.

Below is the relevant code from my project:

1. sign-up-card.tsx:
``` tsx
import React, { useState } from "react";
// Third-party tools/libraries
import { useAuthActions } from "@convex-dev/auth/react";
// Icons
import { TriangleAlert } from "lucide-react";
import { FcGoogle } from "react-icons/fc";
import { FaGithub } from "react-icons/fa";
// Components
import {
  Card,
  CardContent,
  CardDescription,
  CardHeader,
  CardTitle,
} from "@/components/ui/card";
import { Input } from "@/components/ui/input";
import { Button } from "@/components/ui/button";
import { Separator } from "@/components/ui/separator";
// Interface/Types
import { SignInFlow } from "../types";

interface SignUpCardProps {
  setSignType: (signType: SignInFlow) => void;
}

export const SignUpCard = ({ setSignType }: SignUpCardProps) => {
  const { signIn } = useAuthActions();
  const [email, setEmail] = useState(""); // Email
  const [password, setPassword] = useState(""); // Password
  const [confirmPassword, setConfirmPassword] = useState(""); // Confirm Password
  const [pending, setPending] = useState(false); // Page loading state
  const [error, setError] = useState("");

  const handlePasswordSignUp = (e: React.FormEvent<HTMLFormElement>) => {
    e.preventDefault();

    if (password !== confirmPassword) {
      setError("Passwords do not match");
      return;
    }

    setPending(true);
    signIn("password", { email, password, flow: "signup" }).catch((e) =>
      console.log(e)
    );
  };

  return (
    <Card className="w-full h-full p-8">
      <CardHeader className="px-0 pt-0">
        <CardTitle>Start by signing up for an account!</CardTitle>
      </CardHeader>
      <CardDescription className="mb-2">
        Sign up using your email or other methods
      </CardDescription>
      {!!error && (
[...]

PR #194, opened 2025-04-21
added null checks for result properties.

amoscicki:
## Brief description: 
optional chaining added on property accessors, passing errors from proxy, allowing handling them on client side in nextjs

Bunch of folks (including me) were running into issues with error on accessing property on null object. 
#103 
When addressed by adding optional chaining it turned out that when authentication is unsuccessful, the error was the one providing feedback to the client side of the app, and by resolving issue auth errors were only visible on server end. 
Added trivial passing errors in the response from nextjs proxy. This looks like a small can of worms as proper error handling should probably be introduced. 
Since unsure on how maintainers would see best implementation of that, I left the fix at simplest possible implementation that can be built upon by adding matching type declarations on return type of signIn method of the convexAuth function 
https://github.com/get-convex/convex-auth/blob/ffba54c0b3bd9d98ebde75df149eba3de0aafab2/src/server/implementation/index.ts#L403

Errors themselves can look a bit rough as well and properly distinguishing between them is currently a little clunky as requires few splits and manual parsing of the message to get to the actual error type
### example:
`Error: [Request ID: 7a4755d4241579bb] Server Error\nUncaught Error: Could not verify code\n`

----

### DIFF
--- a/src/nextjs/server/proxy.ts
+++ b/src/nextjs/server/proxy.ts
@@ -12,6 +12,7 @@ import {
   logVerbose,
   setAuthCookies,
 } from "./utils.js";
+import { ConvexError } from "convex/values";
 
 export async function proxyAuthActionToConvex(
   request: NextRequest,
@@ -77,7 +78,10 @@ export async function proxyAuthActionToConvex(
       console.error(`Hit error while running \`auth:signIn\`:`);
       console.error(error);
       let response;
-      if (error instanceof Error && error.message) {
+
+      if (error instanceof ConvexError) {
+        response = jsonResponse({ error });
+      } else if (error instanceof Error && error.message) {
         response = jsonResponse({
           error: {
             name: error.name,
@@ -114,7 +118,7 @@ export async function proxyAuthActionToConvex(
       const response = jsonResponse({
         tokens:
           result.tokens !== null
-            ? { token: result.tokens.token, refreshToken: "dummy" }
+            ? { token: result.tokens?.token, refreshToken: "dummy" }
             : null,
       });
       await setAuthCookies(response, result.tokens, cookieConfig);
--- a/src/nextjs/server/proxy.ts
+++ b/src/nextjs/server/proxy.ts
@@ -76,19 +76,32 @@ export async function proxyAuthActionToConvex(
     } catch (error) {
       console.error(`Hit error while running \`auth:signIn\`:`);
       console.error(error);
+      let response;
+      if (error instanceof Error && error.message) {
+        response = jsonResponse({
+          error: {
+            name: error.name,
+            message: error.message,
+            stack: error.stack,
+            cause: error.cause,
+          },
+        });
+      } else {
+        response = jsonResponse({ error });
+      }
       logVerbose(`Clearing auth cookies`, verbose);
-      const response = jsonResponse(null);
       await setAuthCookies(response, null, cookieConfig);
       return response;
     }
-    if (result.redirect !== undefined) {
+    if (result?.redirect !== undefined) {
       const { redirect } = result;
       const response = jsonResponse({ redirect });
       (await getResponseCookies(response, cookieConfig)).verifier =
         result.verifier!;
       logVerbose(`Redirecting to ${redirect}`, verbose);
       return response;
-    } else if (result.tokens !== undefined) {
+    }
+    if (result?.tokens !== undefined) {
       // The server doesn't share the refresh token with the client
       // for added security - the client has to use the server
       // to refresh the access token via cookies.
--- a/src/react/client.tsx
+++ b/src/react/client.tsx
@@ -1,4 +1,4 @@
-import { Value } from "convex/values";
+import { ConvexError, Value } from "convex/values";
 import {
   ReactNode,
   createContext,
@@ -240,7 +240,7 @@ export function AuthProvider({
         { provider, params, verifier },
       );
 
-      if ("error" in result) throw result.error;
+      if ("error" in result) throw new ConvexError(result.error);
 
       if (result?.redirect !== undefined) {
         const url = new URL(result.redirect);
--- a/src/react/client.tsx
+++ b/src/react/client.tsx
@@ -240,7 +240,11 @@ export function AuthProvider({
         { provider, params, verifier },
       );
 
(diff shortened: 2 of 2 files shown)


amoscicki:
I'm pretty certain this is not the intended place, meant to bubble up errors as errors cought there are not instances of ConvexError which I'd guess are the desired ones to be passed through. Happy to built upon this one more if you have any pointers.

ghandic:
Very much requiring something like this as atm I'm not getting any errors thown in nextjs

src/nextjs/server/proxy.ts
```diff
@@ -76,19 +76,32 @@ export async function proxyAuthActionToConvex(
     } catch (error) {
       console.error(`Hit error while running \`auth:signIn\`:`);
       console.error(error);
+      let response;
+      if (error instanceof Error && error.message) {
+        response = jsonResponse({
+          error: {
```
ghandic, line 82: if (error instanceof ConvexError) {
        response = jsonResponse({
          error,
        });
      }

src/react/client.tsx
```diff
@@ -239,15 +239,19 @@ export function AuthProvider({
         "auth:signIn" as unknown as SignInAction,
         { provider, params, verifier },
       );
-      if (result.redirect !== undefined) {
+
+      if ("error" in result) throw new ConvexError(result.error);
```
ghandic, line 243:
```ts
if ("error" in result) throw new ConvexError(result.error.data);
```

ghandic (changes requested):
These needed to be changed for me so I could propagate to the frontend by raising a ConvexError

src/nextjs/server/proxy.ts
```diff
@@ -76,19 +76,32 @@ export async function proxyAuthActionToConvex(
...
+      } else {
+        response = jsonResponse({ error });
+      }
       logVerbose(`Clearing auth cookies`, verbose);
-      const response = jsonResponse(null);
       await setAuthCookies(response, null, cookieConfig);
       return response;
     }
-    if (result.redirect !== undefined) {
+    if (result?.redirect !== undefined) {
```
ghandic, line 96: assuming these dont need modifying too as resposne is returned

src/nextjs/server/proxy.ts
```diff
@@ -76,19 +76,32 @@ export async function proxyAuthActionToConvex(
...
       const { redirect } = result;
       const response = jsonResponse({ redirect });
       (await getResponseCookies(response, cookieConfig)).verifier =
         result.verifier!;
       logVerbose(`Redirecting to ${redirect}`, verbose);
       return response;
-    } else if (result.tokens !== undefined) {
+    }
+    if (result?.tokens !== undefined) {
```
ghandic, line 104: assuming these dont need modifying too as resposne is returned

src/nextjs/server/proxy.ts
amoscicki: Technically no, but if something changes in future and for whatever reason it becomes null it shouldn't hurt to have check for null just in case. Happy to remove it though.

src/react/client.tsx
amoscicki: So if I read it right, idea would then be to only intercept ConvexError types?  otherwise data may not exist on the result.error object.

src/nextjs/server/proxy.ts
amoscicki: idea was to catch any error and the generic Error when stringified to JSON was returning empty object on the frontend in my tests.

amoscicki: When I check using
```typescript
if (error instanceof ConvexError) {
        logVerbose("error detected as ConvexError", verbose);
        response = jsonResponse({ error });
      } else if (error instanceof Error && error.message) {
        logVerbose("error detected as Error", verbose);
        response = jsonResponse({
          error: {
            name: error.name,
            message: error.message,
            stack: error.stack,
            cause: error.cause,
          },
        });
      } else {
        logVerbose("error detected as unknown", verbose);
        response = jsonResponse({ error });
      }
```
      
it logs `[verbose] 2025-04-26T14:38:16.852Z [ConvexAuthNextjs] error detected as Error` for me
I am testing with the OTP provider. Maybe other providers would return it differently?
even more reasons to have it handle all types as this likely shouldn't result in sending user to closes error boundary in nextjs when authentication fails. 
IHMO this should be left for the consumer of the library to handle the way they want.

amoscicki: @ghandic are you ever getting ConvexError instance on your react end? Strangely, even when manually throwing ConvexError inside the `proxy.ts`, on react client side I am getting 
```json
{
    "error": {
        "name": "ConvexError",
        "data": "test"
    },
    "ConvexErrorInstance": false
}
```
```typescript
        console.log({
          error: result.error,
          ConvexErrorInstance: result.error instanceof ConvexError,
        });
```

ghandic:
I'm getting the errors on my frontend, imo I'd prefer to only propagate the errors to the frontend when known and without the full stack trace at it could expose too much detail and parsing it for use is difficult too

ghandic:
I'm also waiting for convex to decide on their pathway either with this or with better auth, because some of the plugins from better auth would be bangin

amoscicki:
> I'm getting the errors on my frontend, imo I'd prefer to only propagate the errors to the frontend when known and without the full stack trace at it could expose too much detail and parsing it for use is difficult too

Sure, the issue is that some of those seem to be the generic Error type which doesn't help. Also for stack trace not to be propagated in nextjs if I'm not mistaken you need to prevent any throws from the backend as by default those would be sent but client side will reroute you to closest error boundary component in your tree.



> I'm also waiting for convex to decide on their pathway either with this or with better auth, because some of the plugins from better auth would be bangin

Yeah I've just seen some new thread appearance on discord. Wonder if that means this impl is to be deprecated?

erquhart:
@amoscicki thanks so much for your work here. I have a smaller scope approach in #208 if you're up for trying it out. Want to fix this in the lowest impact way possible.

erquhart:
@amoscicki good to close?

amoscicki:
Sorry, thought I closed it.

Addressed in https://github.com/get-convex/convex-auth/pull/208

PR #195, opened 2025-04-21
Update passwords.mdx

saranatour1:
according to the comments in file "Password",  you're supposed to add flow to the form in the verification-email, it mentions it in text but doesn't specify it in code. 
following [providers/password](https://github.com/get-convex/convex-auth/blob/ffba54c0b3bd9d98ebde75df149eba3de0aafab2/src/providers/Password.ts#L4-L12)
```ts
 * The `Password` provider supports the following flows, determined
 * by the `flow` parameter:
 *
 * - `"signUp"`: Create a new account with a password.
 * - `"signIn"`: Sign in with an existing account and password.
 * - `"reset"`: Request a password reset.
 * - `"reset-verification"`: Verify a password reset code and change password.
 * - `"email-verification"`: If email verification is enabled and `code` is
 *    included in params, verify an OTP.
```

----

### DIFF
--- a/docs/pages/config/passwords.mdx
+++ b/docs/pages/config/passwords.mdx
@@ -391,7 +391,7 @@ export function SignIn() {
         void signIn("password", formData);
       }}
     >
-      <input name="code" placeholder="Code" type="text" />  
+      <input name="code" placeholder="Code" type="text" />
       <input name="flow" type="hidden" value="email-verification" />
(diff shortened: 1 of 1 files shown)


Nicolapps (approved):
Thank you for your contribution! I tried this flow and I can confirm that adding this is necessary.

PR #200, opened 2025-04-22
Remove Resend version warning

Nicolapps:
Now that Resend 3.5 has been released more than 9 months ago, I think it’s fine just asking people to `npm i resend`.


thomasballinger:
happen to know if resent 3.5 actually works? We'll find out, just curious if you know

Nicolapps:
I’m not sure which version of Resend actually fixed the issue but Resend (which is now at v4.3.0) is definitely working with Convex currently (since @sujayakar is using it in his project)

sujayakar (approved):

PR #191, opened 2025-04-08
Latest version of Jose fails with this example code

webdevcody:
when following the guide, I got this error:

```
webdevcody:~/Workspace/site-sensei$ node generateKeys.mjs
file:///Users/webdevcody/Workspace/site-sensei/node_modules/jose/dist/webapi/lib/asn1.js:20
        throw new TypeError('CryptoKey is not extractable');
              ^

TypeError: CryptoKey is not extractable
    at genericExport (file:///Users/webdevcody/Workspace/site-sensei/node_modules/jose/dist/webapi/lib/asn1.js:20:15)
    at toPKCS8 (file:///Users/webdevcody/Workspace/site-sensei/node_modules/jose/dist/webapi/lib/asn1.js:31:12)
    at exportPKCS8 (file:///Users/webdevcody/Workspace/site-sensei/node_modules/jose/dist/webapi/key/export.js:7:12)
    at file:///Users/webdevcody/Workspace/site-sensei/generateKeys.mjs:4:26

Node.js v20.16.0
```

adding this to the generateKeyPair method seemed to fix it... I'm guessing Jose has a new version.
```
{
  extractable: true,
}
```

so that this script doesn't fail when ran with latest jose version

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #201, opened 2025-04-22
Fix error in manual setup steps for JWT keys

mbokinala:
In the latest version of `jose` (v6), `exportPKCS8()` throws the error `TypeError: CryptoKey is not extractable` unless the option `{ extractable: true }` is set when creating the keypair with `generateKeyPair()`.

The code snippet works when this option is set and Convex auth functions normally.


erquhart:
Thanks for this! Covered in an earlier PR https://github.com/get-convex/convex-auth/pull/191

PR #202, opened 2025-04-22
Migrate from `oslo` to `@oslojs`

Nicolapps:
This removes our usage of the [deprecated](https://www.npmjs.com/package/oslo) `oslo/*` packages and replaces them with their `@oslojs/*` alternatives.

### DIFF
--- a/src/server/implementation/mutations/userOAuth.ts
+++ b/src/server/implementation/mutations/userOAuth.ts
@@ -48,7 +48,7 @@ export async function userOAuthImpl(
     config,
   );
 
-  const code = generateRandomString("0123456789", 8);
+  const code = generateRandomString(8, "0123456789");
   await ctx.db.delete(verifier._id);
   const existingVerificationCode = await ctx.db
     .query("authVerificationCodes")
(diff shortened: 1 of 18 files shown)


src/server/implementation/mutations/userOAuth.ts
```diff
@@ -49,7 +48,7 @@ export async function userOAuthImpl(
     config,
   );
 
-  const code = generateRandomString(8, alphabet("0-9"));
+  const code = generateRandomString(8, "0123456789");
```
Nicolapps, line 51: `@oslojs/crypto` doesn’t have an equivalent `alphabet` function[^1] so I just hardcoded the expected values.

[^1]: See for instance the examples in their documentation: https://crypto.oslojs.dev/examples/random-values

src/server/implementation/utils.ts
```diff
@@ -9,7 +13,17 @@ export function stringToNumber(value: string | undefined) {
 }
 
 export async function sha256(input: string) {
-  return encodeHex(await rawSha256(new TextEncoder().encode(input)));
+  return encodeHexLowerCase(rawSha256(new TextEncoder().encode(input)));
+}
+
+export function generateRandomString(length: number, alphabet: string) {
+  const random: RandomReader = {
```
Nicolapps, line 20: `@oslojs/crypto` now requires the user to explicitly give a source of randomness, I’m creating it using the Web Crypto API as suggested in their documentation: https://crypto.oslojs.dev/examples/random-values

thomasballinger (approved):

PR #205, opened 2025-04-24
Fix error in password reset form

mihajlo-ra92:
`setStep` is defined like this:

```
const [step, setStep] = useState<"forgot" | { email: string }>("forgot");
```
But we try to set the step to `signIn` like this:
```
<button type="button" onClick={() => setStep("signIn")}>
```
And we get this error:
```
Argument of type '"signIn"' is not assignable to parameter of type 'SetStateAction<"forgot" | { email: string; }>'
```

So I just fixed the small mistake by changing `setStep("signIn")` to `setStep("forgot")`.
Hope this helps :)

----


erquhart (approved):
Sorry for the delay - good catch, thanks for this!

PR #209, opened 2025-05-05
Feat/svelte Add support for svelte and sveltekit

mmailaender:
Hi, I tried my best to migrate the React and Next.js functionality to Svelte and SvelteKit.

- there is a README that explains how to use the SvelteKit version
- all playwright tests passed for test-sveltekit
- updated routeMatcher to use newest version of path-to-regexp (affected also nextjs) e.g. `createRouteMatcher(["/product(.*)"])` becomes `createRouteMatcher(["/product{*rest}"])`

I have tested the SvelteKit version as part of the test-sveltekit project, and there it's working. I want to test it now as part of my real project to find more edge cases and test the Svelte version more thoroughly. But I want to create the pull request now as I think it's in a good state to have a discussion, and that you can share your thoughts about it.
I especially think it makes sense to discuss the API as it's not a 100% match, and it is also influenced by the fact that convex-svelte is already branched out from the React client API. 
Also, I noticed that it could make sense to invest some time in deduplicating code for shared code used by React and Svelte (Mutex, routeMatcher, Playwright etc.).

And one small question: During debugging, I recognized that I must create the client with `verbose: true` to activate verbose logging. Did you intend to do it via a prop rather than an env variable? I experience the env variable as quite useful for activating and deactivating verbose logging quite easily, especially if I want to activate it for a short period on deployed instances.

Happy to hear your thoughts and feedback!

----


mmailaender:
@thomasballinger, could you please review and share your feedback? 🙂

thomasballinger:
Whoa, great work here! This will take a bit to review, but it's a good direction. I'll get back to you by tomorrow night with direction here.

Have you tried this in one of your own projects yet?

mmailaender:
> Whoa, great work here! This will take a bit to review, but it's a good direction. I'll get back to you by tomorrow night with direction here.
> 
> Have you tried this in one of your own projects yet?

Yes, I've embedded it in my user and organization management and so far it works well.

mmailaender:
Packaging and installing seem to be a bigger problem. 

## Current Error: 
Sveltekit specific imports in the sveltekit package
```
import { invalidateAll, replaceState } from "$app/navigation";
import { env } from "$env/dynamic/private"
```
throwing errors after being packaged into a tarball and installed. 

But if I install the library via a Symlink, it works fine. As far as I have researched, the problem appears to be that if we bundle our .svelte.ts files into .js, then importing '$app/*' and '$env/*' will break, because these modules only exist within the SvelteKit build pipeline.
One solution seems to be to keep .svelte and .ts files as sources in the published package and let the bundler handle the job on the user side. The `src` folder is already included in the package, but it appears we also need to make some adjustments to the `package.json` and the `tsconfig.json` / `tsconfig.svelte.json` / `tsconfig.sveltekit.json` files, to get it working. Unfortunately, I'm not particularly fond of bundlers, and the last four hours haven't yielded success. Do you have any knowledge that might help on this side? @thomasballinger

thomasballinger:
Yes I can manage this bit! This matches my understanding, that .svelte code can't be bundled. That's fine, it's alright that the packaging process will look different for the svelte-specific bits here. We'll want this to be a separate entry point anyway.

It may be easiest to create a new package for the Svelte auth stuff (which we can package more traditionally for a Svelte library, more similar to https://github.com/get-convex/convex-svelte with lib etc.) and we expose the parts of convex-auth that are necessary in that code. I haven't had a change to look into this yet, hoping to by tomorrow morning. But without looking at the bundling issues specifically that's my bet for the smoothest path forward: using Convex Auth with Svelte would require installing two packages, @convex-dev/auth and @convex-dev/svelte-auth.

mmailaender:
@thomasballinger have you already had a chance to work on it? If not, I would give this another try in the next days and put Svelte and Svelte Kit into a separate package.

mmailaender:
@thomasballinger I finally got the bundler working. I have created an own version that only has the Svelte and SvelteKit versions included and uses the server part from @convex-dev/auth.

https://github.com/mmailaender/convex-svelte-auth

It would be great to get your feedback and your thoughts on BetterAuth. 🚀

BraveDevelopment:
> @thomasballinger I finally got the bundler working. Based on your feedback, I have created an own version that only has the Svelte and SvelteKit versions included and imports the server part from @convex-dev/auth.
> 
> https://github.com/mmailaender/convex-svelte-auth
> 
> It would be great to get your feedback and your thoughts on BetterAuth. 🚀

Nice work! 
Any news on when this will get integrated? @thomasballinger

PR #208, opened 2025-05-04
match Convex Action error handling in nextjs api proxy

erquhart:
Fixes #103

Smaller scope alternative to #194

When Convex Auth runs from a react client (non-server), the sign in action throws on invalid input, eg., missing or invalid password. The nextjs auth api proxy swallows the error and returns a 200 response with a null body.

This solution:
- Returns a 400 with an `error` key in the body containing the error message (complete with stack trace, mirroring the non-next react behavior)
- Parses and throws the error in the nextjs authenticated function, which is only used for signIn and signOut
- Avoids changes outside of the immediately affected code paths (no react client changes)

Should we be throwing a stack trace? Maybe not. But the current nextjs behavior is a bug, and this solution matches what we already do intentionally for non nextjs apps.

Recommend improvements to this approach be split to a separate issue, the only proposal here is fixing the bug and aligning with existing working behavior.

----

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #210, opened 2025-05-07
require error message from backend in test

erquhart:
Adds an implicit check to an existing test, ensures backend error message propagates to Next.js client.

Validates #208 

----


erquhart:
Bah accidentally pushed with version update

PR #216, opened 2025-05-15
Increase window location check specificity for Expo 53+

### LINKED ISSUE #206: NotSupportedError: Cannot set "location.href"
In this code, `window.location !== undefined` is used to check if react native is actually running the code. 

But this check is not false on react native leading to this error when trying to set window.location.href value: `NotSupportedError: Cannot set "location.href"`

I believe instead trying to check if we are actually in react native, this code part can be completly removed and let the caller handle the redirection ( or pass a boolean param instead of automatic checking )

I can do a PR if someone make the choose

```ts
const signIn = useCallback(
    async (provider?: string, args?: FormData | Record<string, Value>) => {
      const params =
        args instanceof FormData
          ? Array.from(args.entries()).reduce(
              (acc, [key, value]) => {
                acc[key] = value as string;
                return acc;
              },
              {} as Record<string, string>,
            )
          : args ?? {};

      const verifier = (await storageGet(VERIFIER_STORAGE_KEY)) ?? undefined;
      await storageRemove(VERIFIER_STORAGE_KEY);
      const result = await client.authenticatedCall(
        "auth:signIn" as unknown as SignInAction,
        { provider, params, verifier },
      );
      if (result.redirect !== undefined) {
        const url = new URL(result.redirect);
        await storageSet(VERIFIER_STORAGE_KEY, result.verifier!);
        // Do not redirect in React Native
        if (window.location !== undefined) {
          window.location.href = url.toString();
        }
        return { signingIn: false, redirect: url };
      } else if (result.tokens !== undefined) {
        const { tokens } = result;
        logVerbose(`signed in and got tokens, is null: ${tokens === null}`);
        await setToken({ shouldStore: true, tokens });
        return { signingIn: result.tokens !== null };
      }
      return { signingIn: false };
    },
    [client, setToken, storageGet],
  );
```

packages.json
```json
{
  "name": "redacted",
  "main": "expo-router/entry",
  "version": "1.0.0",
  "scripts": {
    "start": "expo start",
    "reset-project": "node ./scripts/reset-project.js",
    "android": "expo start --android",
    "ios": "expo start --ios",
    "web": "expo start --web",
    "lint": "expo lint"
  },
  "dependencies": {
    "@auth/core": "0.39.0",
    "@convex-dev/auth": "^0.0.82",
    "@expo/vector-icons": "^14.1.0",
    "@react-navigation/bottom-tabs": "^7.3.10",
    "@react-navigation/elements": "^2.3.8",
    "@react-navigation/native": "^7.1.6",
    "@rn-primitives/accordion": "^1.1.0",
    "@rn-primitives/alert-dialog": "^1.1.0",
    "@rn-primitives/aspect-ratio": "^1.1.0",
    "@rn-primitives/avatar": "^1.1.0",
    "@rn-primitives/checkbox": "^1.1.0",
    "@rn-primitives/collapsible": "^1.1.0",
    "@rn-primitives/dialog": "^1.1.0",
    "@rn-primitives/dropdown-menu": "^1.1.0",
    "@rn-primitives/label": "^1.1.0",
    "@rn-primitives/menubar": "^1.1.0",
    "@rn-primitives/navigation-menu": "^1.1.0",
    "@rn-primitives/popover": "^1.1.0",
    "@rn-primitives/progress": "^1.1.0",
    "@rn-primitives/radio-group": "^1.1.0",
    "@rn-primitives/select": "^1.1.0",
    "@rn-primitives/separator": "^1.1.0",
    "@rn-primitives/slot": "^1.1.0",
    "@rn-primitives/switch": "^1.1.0",
    "@rn-primitives/tabs": "^1.1.0",
    "@rn-primitives/toggle": "^1.1.0",
    "@rn-primitives/toggle-group": "^1.1.0",
    "@rn-primitives/types": "^1.1.0",
    "class-variance-authority": "^0.7.1",
    "clsx": "^2.1.1",
    "convex": "^1.24.0",
    "expo": "~53.0.6",
    "expo-auth-session": "~6.1.5",
    "expo-blur": "~14.1.4",
    "expo-constants": "~17.1.5",
    "expo-font": "~13.3.1",
    "expo-haptics": "~14.1.4",
    "expo-image": "~2.1.6",
    "expo-linking": "~7.1.4",
    "expo-router": "~5.0.5",
    "expo-secure-store": "~14.2.3",
    "expo-splash-screen": "~0.30.8",
    "expo-status-bar": "~2.2.3",
    "expo-symbols": "~0.4.4",
    "expo-system-ui": "~5.0.7",
    "expo-web-browser": "~14.1.6",
    "lucide-react-native": "^0.507.0",
    "nativewind": "^4.1.23",
    "oslo": "^1.2.1",
    "react": "19.0.0",
    "react-dom": "19.0.0",
    "react-native": "0.79.2",
    "react-native-gesture-handler": "~2.24.0",
    "react-native-reanimated": "~3.17.4",
    "react-native-safe-area-context": "5.4.0",
    "react-native-screens": "~4.10.0",
    "react-native-svg": "15.11.2",
    "react-native-web": "~0.20.0",
    "react-native-webview": "13.13.5",
    "resend": "^4.5.0",
    "sonner": "^2.0.3",
    "sonner-native": "^0.19.0",
    "tailwind-merge": "^3.2.0",
    "tailwindcss": "^3.4.17",
    "tailwindcss-animate": "^1.0.7"
  },
  "devDependencies": {
    "@babel/core": "^7.25.2",
    "@types/react": "~19.0.10",
    "typescript": "~5.8.3",
    "eslint": "^9.25.0",
    "eslint-config-expo": "~9.2.0"
  },
  "private": true
}
```

erquhart:
Fixes #206

Expo runtime introduced a default polyfill to metro runtime, making `window.location` no longer undefined. This breaks some of our client checks for the RN environment.

This fix makes checks more specific and looks for the actual property we're interacting with to be undefined.

#207 addresses this by adding an explicit prop, which definitely works, but I'd like to avoid adding more props. The library should just handle this for users.

A patch in #206 used `navigator.product` as a reference, which does appear to be reliably set to `ReactNative` ([source](https://github.com/facebook/react-native/blob/ca764bb5115d2e7baa6e07c49e64d4806f5b3d4a/packages/react-native/Libraries/Core/setUpNavigator.js#L18)). This is an option for the future if more changes in platforms make checking location object properties a dubious practice.

### DIFF
--- a/src/react/client.tsx
+++ b/src/react/client.tsx
@@ -243,7 +243,7 @@ export function AuthProvider({
         const url = new URL(result.redirect);
         await storageSet(VERIFIER_STORAGE_KEY, result.verifier!);
         // Do not redirect in React Native
-        if (window.location !== undefined) {
+        if (window.location?.href !== undefined) {
           window.location.href = url.toString();
         }
         return { signingIn: false, redirect: url };
@@ -359,7 +359,7 @@ export function AuthProvider({
         return;
       }
       const code =
-        typeof window?.location !== "undefined"
+        typeof window?.location?.search !== "undefined"
           ? new URLSearchParams(window.location.search).get("code")
           : null;
       // code from URL is only consumed initially,


thomasballinger (approved):

thomasballinger:
If there's an expo-endorsed way to detect the environment even better, but let's do this

PR #265, opened 2025-12-04
Require authorize field in email provider config

erquhart:
The `authorize` field in a custom Email provider config can be skipped unintentionally as it isn't required in types. Not providing an authorize function must be allowed to support "magic link" auth. This change:
- Makes the field required, but still accepts an explicit undefined so the opt-out is more clear.
- Adds an explanatory docstring

Unrelated maintenance:
- adds release scripts

## Summary by CodeRabbit

* **Chores**
  * Added new release management scripts for pre-release versioning and publishing workflows.
  * Streamlined prepublish build process to focus on compilation only.
  * Updated build automation tool dependencies.
  * Clarified API type definitions for email configuration parameters.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,5 @@ next-env.d.ts
 
 playwright-report/
 test-results/
+
+*.tgz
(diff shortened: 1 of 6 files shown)


erquhart:
The core gap this is meant to fill should be addressed through #285.

Issue #276, opened 2026-01-10
Bug: window is undefined


nyoung697:
These two functions throw errors in non-browser environments (react-native) where the `window` object is undefined:

```
function browserAddEventListener<K extends keyof WindowEventMap>(
  type: K,
  listener: (this: Window, ev: WindowEventMap[K]) => any,
  options?: boolean | AddEventListenerOptions,
): void {
  window.addEventListener?.(type, listener, options);
}

function browserRemoveEventListener<K extends keyof WindowEventMap>(
  type: K,
  listener: (this: Window, ev: WindowEventMap[K]) => any,
  options?: boolean | EventListenerOptions,
): void {
  window.removeEventListener?.(type, listener, options);
}
```

Temporary workaround was to create a shim:

```
// src/shims/window.ts
export function installWindowShim() {
  // Only in React Native / non-browser environments
  if (typeof globalThis.window === "undefined") {
    // Make it exist
    (globalThis as any).window = {};
  }

  // Make addEventListener "safe" if the library optional-chains it.
  // (Your code uses window.addEventListener?. so {} is enough,
  // but some libs call window.addEventListener without ?.)
  if (typeof (globalThis as any).window.addEventListener !== "function") {
    (globalThis as any).window.addEventListener = () => {};
  }
  if (typeof (globalThis as any).window.removeEventListener !== "function") {
    (globalThis as any).window.removeEventListener = () => {};
  }
}
```

PR #277, opened 2026-01-10
Fix window is undefined bug

nyoung697:
closes: https://github.com/get-convex/convex-auth/issues/276

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved runtime errors that could occur in server-side rendering and other non-browser environments.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


erquhart (approved):
Makes sense 👍

PR #280, opened 2026-02-09
Add beforeSessionCreation callback hook

focus-stefano:
### Problem

The `createOrUpdateUser` callback only fires during account creation (sign-up, OAuth, email/phone OTP). The credentials sign-in flow bypasses it entirely (`retrieveAccount` → `callSignIn` → `createSession`), so there's no hook to reject a returning user's login (e.g. banned users).

### Solution

New optional `beforeSessionCreation` callback that runs in `createSession()` before the session is persisted. Covers all sign-in flows. Throw to reject.

### Changes

- `src/server/types.ts` — callback type + JSDoc
- `src/server/implementation/sessions.ts` — 3-line guard in `createSession()`

### Usage

```ts
callbacks: {
  async beforeSessionCreation(ctx, { userId }) {
    const user = await ctx.db.get(userId);
    if (user?.banned) throw new Error("Account is banned");
  },
}
```

----

## Summary by CodeRabbit

* **New Features**
  * Added optional pre-session creation callback that runs during all sign-in flows, enabling custom validation logic and the ability to reject sign-ins.

### DIFF
--- a/src/server/implementation/sessions.ts
+++ b/src/server/implementation/sessions.ts
@@ -89,6 +89,9 @@ async function createSession(
   userId: GenericId<"users">,
   config: ConvexAuthConfig,
 ) {
+  if (config.callbacks?.beforeSessionCreation !== undefined) {
+    await config.callbacks.beforeSessionCreation(ctx, { userId });
+  }
   const expirationTime =
     Date.now() +
     (config.session?.totalDurationMs ??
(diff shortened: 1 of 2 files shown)


erquhart:
Looks good 👍 - replacing with #296 (maintainer edits were disabled for this PR)

PR #305, opened 2026-04-04
fix: handle Microsoft multi-tenant issuer mismatch

moshemarciano:
## Problem

When using `MicrosoftEntraId` with the default multi-tenant ("common") endpoint, the OAuth callback fails. OIDC discovery returns `issuer: .../common/v2.0`, but ID tokens carry tenant-specific `iss` claims (`.../{tenantId}/v2.0`). `oauth4webapi`'s issuer validation rejects these tokens.

## Fix

Use `oauth4webapi`'s exported `_expectedIssuer` Symbol extension point on the discovered `AuthorizationServer` for `microsoft-entra-id` providers. This accepts the token's own `iss` claim while still validating the JWT signature — the same approach `@auth/core` (Auth.js) uses internally for Microsoft.

The fix only applies in the discovery code path (when `authorization`, `token`, or `userinfo` endpoints are not all explicitly provided), which is the common configuration for Microsoft Entra ID.

## Testing

- Multi-tenant Microsoft sign-in now succeeds (previously failed with issuer mismatch)
- Single-tenant (`tenantId: "..."`) continues to work — discovery returns the tenant-specific issuer which matches tokens
- Other OIDC providers unaffected — fix is scoped to `config.id === "microsoft-entra-id"`

Closes #304

## Summary by CodeRabbit

* **New Features**
  * Improved Microsoft Entra ID (Azure AD) OIDC sign-in: issuer matching now respects the token's reported issuer for multi-tenant alias endpoints.

* **Bug Fixes**
  * Fixed validation failures for sign-ins using common/organizations/consumers multi-tenant endpoints, reducing incorrect or failed sign-in validations.

### DIFF
--- a/src/server/oauth/convexAuth.ts
+++ b/src/server/oauth/convexAuth.ts
@@ -89,9 +89,17 @@ export const defaultCookiesOptions: (
 
 // Microsoft multi-tenant: aliases like "common", "organizations", "consumers"
 // discover issuer .../{alias}/v2.0 but tokens carry tenant-specific
-// iss claims (.../{tenantId}/v2.0). Use oauth4webapi's
-// _expectedIssuer extension to accept the token's own iss,
-// matching the approach used by @auth/core for Microsoft providers.
+// iss claims (.../{tenantId}/v2.0).
+//
+// oauth4webapi's processDiscoveryResponse validates that the ID token's
+// iss matches the discovered issuer, which fails for these aliases.
+// The library exports a _expectedIssuer Symbol as an extension point
+// specifically for this case — setting it to a function replaces the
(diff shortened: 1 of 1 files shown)


aissamChia:
We hit the same issue while integrating Microsoft Entra ID with Convex Auth for multi-tenant sign-in (`common` endpoint + personal Microsoft accounts).

After Microsoft consent, the callback failed with:

`unexpected JWT "iss" (issuer) claim value`

We applied this patch locally via `patch-package` and confirmed the fix works end-to-end: token exchange completes, the user is created in Convex, and sign-in succeeds.

Thanks for putting this together — would love to see it land in a release.

PR #306, opened 2026-04-06
Pass authUserId to createOrUpdateUser for secure account linking

rorhug:
## Problem

In `createOrUpdateUser` it's not currently possible to know who the currently logged in user is.

Given an app with Github auth and Google auth, without customisation if you first log in with Github and then Google, the Google signin creates a second user and the session becomes that. Therefore we need custom account linking logic. 

Current docs for account linking: https://labs.convex.dev/auth/advanced#controlling-user-creation-and-account-linking-behavior

However, although the docs recommend using `createOrUpdateUser(ctx, args)` the ctx does not have the current session because it's being ran from a convex HTTP `*.site` URL which doesn't have the cookies. Therefore the following code prints "my user id is null" while having clicked "connect Google" even though we are logged in already.

```ts
export const { auth, signIn, signOut, store, isAuthenticated } = convexAuth({
  providers: [
    GitHubProvider(),
    GoogleProvider(),
  ],
  callbacks: {
    async createOrUpdateUser(ctx, args) {
      const userId = await getAuthUserId(ctx);
      console.log("my user id is ", userId)
    }
  }
})
```

Email based account linking is not good enough.

- it's insecure
- also people often have accounts they want to link that have different emails
- some services don't provide user email in the OAuth profile [for example ;)](https://docs.convex.dev/platform-apis/oauth-applications)

## Solution

- Pass `authUserId` to the `createOrUpdateUser` callback when an OAuth flow is started while another user is already signed in.
- Update the advanced docs example to show how to use `authUserId` for OAuth account linking, fix a small typo in the callback docs.

## Why

Custom account-linking logic needs access to the currently authenticated user during signed-in OAuth flows. Without that, `createOrUpdateUser` callbacks can't reliably attach a new OAuth account to the existing logged in user who initiated the flow.

- `authUserId` is distinct from `existingUserId`
  - `existingUserId: Id<"users"> | null` refers to the user that may already be linked to the account being processed
  - `authUserId: Id<"users"> | null` **NEW** refers to the user who may currently be logged in.

## Test plan

- [ ] Verify the `createOrUpdateUser` callback receives `authUserId` during an OAuth flow started while signed in
- [ ] Verify `authUserId` is `null` for flows without an authenticated session

----

### DIFF
--- a/docs/pages/advanced.mdx
+++ b/docs/pages/advanced.mdx
@@ -89,6 +89,12 @@ export const { auth, signIn, signOut, store, isAuthenticated } = {
         return args.existingUserId;
       }
 
+      if (args.authUserId) {
+        // During OAuth account linking, attach the new account to the
+        // currently authenticated user who started the flow.
+        return args.authUserId;
+      }
+
       // Implement your own account linking logic:
       const existingUser = await findUserByEmail(ctx, args.profile.email);
       if (existingUser) return existingUser._id;
@@ -101,6 +107,10 @@ export const { auth, signIn, signOut, store, isAuthenticated } = {
   },
 };
 ```
+
+`args.existingUserId` is the user already linked to the account being signed in,
+if one exists. `args.authUserId` identifies the currently authenticated user so
+you can link the new OAuth account to them. This is useful for OAuth flows started while signed in.
 
 When you provide this callback, the library doesn't create or update users at
 all. It is up to you to implement all the necessary logic for all providers you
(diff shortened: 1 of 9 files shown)


rorhug:
Hey @erquhart, could you take a look at this?

erquhart:
@rorhug thanks for this! It would definitely be good to better enable account linking, and the approach looks viable, but I would need to take some time ensuring its fully secure. Given the maintenance mode status of Convex Auth that's unlikely to happen immediately, but I want to leave this open if that's alright for future consideration.

PR #326, opened 2026-04-29
Export useAuth hook for custom auth providers

Hades32:
Allow consumers to build their own auth providers.

We need that because our frontend is connecting to two different Convex instances, where one of them is the auth provider.

----


erquhart:
This is no problem to do, useAuth api has to match the provider contract anyway, so not meaningfully broadening our committed api.

erquhart (approved):

PR #336, opened 2026-06-01
Downgrade log levels for expected auth flows

Hades32:
Several normal, client-driven auth conditions were logged at ERROR (e.g. a stale frontend refreshing an expired session). Downgrade expected sign-out / sign-in failures to INFO and keep genuine anomalies (token reuse, rate limiting, tampering) at WARN.

This matches the best practice for log levels, and allows downstream users to use log-levels for alerting. e.g. we only log errors for technical issues a developer needs to look at, and warns for other unexpected behavior.

----


erquhart (approved):
Good changes, thanks for this

PR #354, opened 2026-06-22
Prep work for next version of Convex Auth

dowski:
Includes:

 * Convex AI files
 * Prototype subrepo
 * Custom AGENTS.md guidance for the project

----

### DIFF
--- a/.agents/skills/convex-setup-auth/SKILL.md
+++ b/.agents/skills/convex-setup-auth/SKILL.md
@@ -1,0 +1,187 @@
+---
+name: convex-setup-auth
+description:
+  Sets up Convex auth, identity mapping, and access control. Use for login, auth
+  providers, users tables, protected functions, or roles in a Convex app.
+---
+
+# Convex Authentication Setup
+
+Implement secure authentication in Convex with user management and access
+control.
+
+## When to Use
+
+- Setting up authentication for the first time
+- Implementing user management (users table, identity mapping)
+- Creating authentication helper functions
+- Setting up auth providers (Convex Auth, Clerk, WorkOS AuthKit, Auth0, custom
+  JWT)
+
+## When Not to Use
+
+- Auth for a non-Convex backend
+- Pure OAuth/OIDC documentation without a Convex implementation
+- Debugging unrelated bugs that happen to surface near auth code
+- The auth provider is already fully configured and the user only needs a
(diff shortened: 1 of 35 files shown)


.agents/skills/convex-setup-auth/SKILL.md
```diff
@@ -0,0 +1,187 @@
+---
```
Nicolapps, line 1: I guess that this is the base skill referring to pre-v2 auth, right?

AGENTS.md
```diff
@@ -0,0 +1,27 @@
+This is Convex Auth, an authentication framework built on top of Convex.
+
+Because it's a framework, some guidelines that are specific to building apps might not apply. However, much of the guidance you'll receive about Convex development is spot on.
+
+Just watch out for things like the location of the `convex/` folder and other stuff that might vary between a framework that is meant to be used with a Convex app and a Convex app itself.
+
+This is actually a new start for Convex Auth - commit b6cf31f06201062cbd86687f5da1a1434dcb660c marks the (likely) last checkpoint of the old system. You may or may not see the old system in the codebase, depending on what stage of development we're at. You can reference files at that commit if you want to look back at the old system.
```
Nicolapps, line 7: I wonder if this PR should completely strip out the old code to avoid this issue

prototype
Nicolapps, line 1: Is it intentional that this is a submodule?

erquhart (commented):
Makes sense, curious on the subrepo approach

.agents/skills/convex-setup-auth/SKILL.md
dowski: This seems to be about using auth with Convex generally (Convex Auth, Auth0, Clerk, etc).

AGENTS.md
dowski: I think it will be useful to let LLMs refer to the existing code to validate/compare approaches. We could do it w/ a subrepo like the prototype, but let's start with this.

prototype
dowski: Yes, this is for LLMs to refer to the prototype during development.

dowski:
> Makes sense, curious on the subrepo approach

Just for allowing LLMs to refer to the prototype during development.

erquhart (approved):

PR #356, opened 2026-06-24
Add core auth component

dowski:
The core auth component that owns sessions and accounts. Offers integration with applications via the `setupCore` function which accepts an app provided `createOrUpdateUser` function reference.

I suggest focused review on the `src/components` and `src/lib` additions (minus generated files) and the `generate-auth-keys.mjs` script. That'll be ~1000 lines vs the 1800+ listed here on the PR.

In particular, pay attention to how the `AuthClaims` and `TokenBundle` types are used. `AuthClaims` is how providers communicate account data back to the core, and `TokenBundle` is what will get passed to (eventual) client code.

----


dowski:
@erquhart and @Nicolapps - I added you both as reviewers since this is a piece we'll all be building on. I think it'll be good to have familiarity with it, plus your feedback.

.eslintrc.cjs
```diff
@@ -10,6 +10,10 @@ module.exports = {
     "test",
     "docs",
     "dist",
+    "prototype",
+    "examples",
+    "scripts",
```
Nicolapps: nit: can we include all files in ESLint?

Also, if we want to continue building in this repo, we’ll need to update ESLint to a supported version

scripts/generate-auth-keys.mjs
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env node
```
Nicolapps, line 1: I can’t wait to get rid of this file 😃

examples/react-minimal/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "@convex-dev/auth-example-react-minimal",
+  "version": "0.0.0",
+  "private": true,
+  "type": "module",
+  "dependencies": {
+    "@convex-dev/auth": "file:../..",
+    "convex": "^1.40.0",
+    "react": "^18.3.1",
+    "react-dom": "^18.3.1"
```
Nicolapps: nit: would you mind making sure we use the latest versions of convex/React/Vite/TypeScript in the new code, so that they are already up to date?

src/components/core/http.ts
```diff
@@ -0,0 +1,25 @@
...
+
+// The core serves its own JWKS. AUTH_JWKS is a static env var set by running
+// `npx generate-auth-keys` and is served verbatim here. Convex mounts these
+// routes under the prefix the app declares in convex.config.ts
+// (`app.use(core, { httpPrefix: "/auth" })`), so the served path is
+// <prefix>/.well-known/jwks.json — point auth.config.ts's `jwks` URL there.
+const http = httpRouter();
+
+http.route({
+  path: "/.well-known/jwks.json",
```
Nicolapps, line 12: I double-checked this: the OIDC allows users to mount `.well-known` URLs at a non-root path (as long as the `ISSUER` contains the additional path components).

Linking to my research: https://app.notion.com/p/Convex-Auth-v2-Auth-session-storage-JWTs-vs-session-strings-382b57ff32ab804593ccf5b024d97df3?d=38ab57ff32ab803497a8001c4d1f06a4&source=copy_link#383b57ff32ab80cf9e9ad618cb1b6ae8

src/components/core/public.ts
```diff
@@ -0,0 +1,314 @@
...
+ * on *every* sign-in: with no `userId` the first time (it mints/returns the
+ * user id), and with the known `userId` on every return (so it can sync the
+ * user record from the latest claims). The profile the core holds for the
+ * account is refreshed to match. Returns just what minting a session needs: the
+ * account id and its app user id.
+ */
+async function resolveAccount(
+  ctx: MutationCtx,
+  claims: AuthClaims,
+  createOrUpdateUserHandle: string,
```
Nicolapps: nit: can you please type this as `FunctionHandle<…>` and do the cast in `signIn`?

erquhart (approved):
Overall makes sense, the one thing we'll want to do hardening on is refresh token handling. I don't think that needs to block this PR.

src/components/core/public.ts
```diff
@@ -0,0 +1,314 @@
...
+/** Revoke a session (sign out). Idempotent. */
+export const signOut = mutation({
+  args: { refreshToken: v.string() },
+  returns: v.null(),
+  handler: async (ctx, args) => {
+    const session = await sessionByHash(
+      ctx,
+      await hashToken(args.refreshToken),
+    );
+    if (session) await ctx.db.delete(session._id);
```
Nicolapps: The code in this file uses the old `db.*` APIs without table names.

Can you please make sure that the latest version of the [Convex ESLint plugin](https://docs.convex.dev/eslint) is installed with all recommended rules?

This will make sure we won’t have security issues with malicious users intentionally providing IDs for the wrong table in places where the table ID might not be checked at runtime. It also makes the code more future-proof

src/lib/types.ts
```diff
@@ -0,0 +1,71 @@
...
+export type CreateOrUpdateUserFn = FunctionReference<
+  "mutation",
+  "internal",
+  {
+    provider: string;
+    providerAccountId: string;
+    profile: Record<string, unknown>;
+    userId?: string;
+  },
+  string
```
Nicolapps: If we go for the `createOrUpdateUser` approach, I strongly think that we need to type the return value as `Id<"users">`, not as an arbitrary string.

The current pattern allows `handler: async (_ctx, args) => args.userId ?? args.providerAccountId`, but I think that this can be dangerous. In particular, if we ever support a provider where the attacker might have some control over their user ID, it might allow them to confuse the auth system into giving them control of an account.

I understand that it can be convenient to be able to skip the `users` table when it’s not needed, but I’d rather have a clear mental model where Convex apps should only ever use a single string to refer to a user, which would be a `Id<"users">`. Even in apps where there is a single login method that provides its own account IDs, using custom user IDs could turn out to be annoying if the app needs to support another log in method later.

src/components/core/core.test.ts
```diff
@@ -0,0 +1,314 @@
+/// <reference types="vite/client" />
```
Nicolapps: Thank you for including tests!

src/components/core/public.ts
```diff
@@ -0,0 +1,314 @@
...
+
+  const account = await accountByIdentity(
+    ctx,
+    claims.provider,
+    claims.providerAccountId,
+  );
+  if (account) {
+    // Returning identity: hand the app the latest claims with the known user id
+    // so it can update its own user record, then refresh the profile we hold.
+    await ctx.runMutation(createOrUpdateUser, {
```
Nicolapps: If I understand correctly, in this case we expect `createOrUpdateUser` to always return `acocunt.userId`. Can we please check this at run time, and throw an internal error (`new Error`) when it is not the case?

src/lib/types.ts
```diff
@@ -0,0 +1,71 @@
...
+ * paths can simply ignore `userId`.
+ */
+export type CreateOrUpdateUserFn = FunctionReference<
+  "mutation",
+  "internal",
+  {
+    provider: string;
+    providerAccountId: string;
+    profile: Record<string, unknown>;
+    userId?: string;
```
Nicolapps: nit: Can we change this to `userId: string | null`? It sounds semantically important

src/lib/types.ts
```diff
@@ -0,0 +1,71 @@
...
+ * provider profile onto that user). Apps that don't need to react to the update
+ * paths can simply ignore `userId`.
+ */
+export type CreateOrUpdateUserFn = FunctionReference<
+  "mutation",
+  "internal",
+  {
+    provider: string;
+    providerAccountId: string;
+    profile: Record<string, unknown>;
```
Nicolapps: It’s really unfortunate that this can’t be strongly typed 😔

But I don’t think that we can really do it better under the current design. Since `setupCore` is called before the providers are set up, it doesn’t know which providers are available.

If we decide to ship this design, we’ll probably need to instruct users to have a select over `provider`, make an explicit cast in each case, and throw if they encounter an unexpected provider.

src/lib/types.ts
```diff
@@ -0,0 +1,71 @@
...
+ * what the core expects. The app keeps ownership of its users table — the core
+ * only holds a reference to this one mutation.
+ *
+ * It serves every auth path: sign-in calls it with no `userId` the first time
+ * an identity is seen (create the user, return the id) and with the resolved
+ * `userId` on every later sign-in (update the user from the latest profile);
+ * `linkToCurrentUser` calls it with the already-known `userId` (sync the
+ * provider profile onto that user). Apps that don't need to react to the update
+ * paths can simply ignore `userId`.
+ */
```
Nicolapps, line 77: I think that the documentation of `CreateOrUpdateUserFn` needs to be more detailed. This function is the central point where users can customize the auth behavior, so it’s important that we explain clearly what we expect them to do.

In particular, I think the documentation should explain clearly that:
- If they receive a user ID, they _can_ update app-level data to save the data that they might need (e.g. Google account details), and then they need to return the same value
- If they do _not_ receive a user ID, they need to either create a row in `users` and return its ID, or return the ID of an existing `users` row (e.g. to link a new GitHub account with a verified email address with an existing email+password account that has the same address).
- This function is allowed to throw `ConvexError`s to perform validation (e.g. if we want to have OAuth log in, but only allow `@convex.dev` email addresses).
  - I’d love for the errors thrown in this project to be typesafe, because it would make it easier to localize in the frontend. But I’m not sure how we can achieve this under this design.

src/components/core/crypto.ts
```diff
@@ -0,0 +1,49 @@
...
+  subject: string;
+  issuer: string;
+  audience: string;
+  expiresInSeconds: number;
+  claims?: Record<string, unknown>;
+}): Promise<{ token: string; expiresAt: number }> {
+  const key = await importPKCS8(opts.privateKeyPkcs8, ALG);
+  const nowSeconds = Math.floor(Date.now() / 1000);
+  const expSeconds = nowSeconds + opts.expiresInSeconds;
+  const token = await new SignJWT({ ...opts.claims })
```
Nicolapps: What is the `opts.claims` for? It looks unused at the moment.

This looks a little bit dangerous since some claims (like `iss`, `exp` and `sub`) have special meanings.

src/components/core/public.ts
```diff
@@ -0,0 +1,314 @@
...
+  }
+  if (accessTokenTtlSeconds >= refreshTokenTtlSeconds) {
+    throw new Error(
+      "Access-token TTL must be shorter than the refresh-token TTL.",
+    );
+  }
+  return { accessTokenTtlSeconds, refreshTokenTtlSeconds };
+}
+
+async function mintAccessToken(
```
Nicolapps, line 91: My understanding is that we’ll probably switch to sessions before the release of Auth v2 (see https://app.notion.com/p/convex-dev/Convex-Auth-v2-Auth-session-storage-JWTs-vs-session-strings-382b57ff32ab804593ccf5b024d97df3?source=copy_link). Could you please add a TODO comment at the top of the file indicating that this implementation is temporary?

src/components/core/schema.ts
```diff
@@ -0,0 +1,36 @@
...
+import { v } from "convex/values";
+
+export default defineSchema({
+  // Maps a provider-scoped identity to an opaque app user id. The app owns the
+  // actual users table; we only store the id string it gives us back.
+  accounts: defineTable({
+    provider: v.string(),
+    providerAccountId: v.string(),
+    userId: v.string(),
+    profile: v.any(),
```
Nicolapps: Why do we store the `profile` data here?

I don’t think that it makes sense to store it here because:
- In `createOrUpdateUser` the app can access the profile, so it can already store this information if necessary
- We currently don’t have a way for the app to read `accounts.profile`
- It might be that the app receives some information from the auth provider that it genuinely doesn’t want to store long term (e.g. for privacy reasons)
- `profile` would need to be typed as `v.any()`, which is annoying to work with (there is no type safety)

src/components/core/setup.ts
```diff
@@ -0,0 +1,133 @@
...
+ * and (since the client refreshes shortly before expiry) comfortably longer
+ * than a few seconds.
+ */
+export function setupCore(opts: {
+  component: ComponentApi;
+  createOrUpdateUser: CreateOrUpdateUserFn;
+  /** Access-token lifetime in seconds. Defaults to 60 (1 minute). */
+  accessTokenTtlSeconds?: number;
+  /** Refresh-token lifetime in seconds. Defaults to 30 days. */
+  refreshTokenTtlSeconds?: number;
```
Nicolapps: When going with a framework-y approach, I think that we will need to document clearly what happens if users change parameters after publishing their app first. For instance, here users might assume that by reducing `accessTokenTtlSeconds` it’s going to change the TTL of existing access tokens, while in practice it’s only going to affect new ones.

src/components/core/setup.ts
```diff
@@ -0,0 +1,133 @@
...
+  const completeSignIn = async (
+    ctx: GenericActionCtx<any>,
+    claims: AuthClaims,
+  ): Promise<TokenBundle> => {
+    const createOrUpdateUserHandle =
+      await createFunctionHandle(createOrUpdateUser);
+    return await ctx.runMutation(component.public.signIn, {
+      claims,
+      createOrUpdateUserHandle,
+      issuer: issuer(),
```
Nicolapps, line 97: nit: to conform to the OIDC spec, the issuer would need to include the path components (including the HTTP prefix)

No need to fix this is we plan to switch to sessions, but maybe we can add a FIXME comment for clarity

src/components/core/schema.ts
```diff
@@ -0,0 +1,36 @@
...
+  })
```
Nicolapps, line 11: Having an `accounts` table in the core component is a choice that doesn’t seem obvious. I wonder if we’ll need to revisit this choice in the future before releasing Auth v2. I would love to review or write a design doc about this choice, comparing it with the alternatives (e.g. standalone accounts component, or storing the account associations in userspace).

A few initial thoughts about this design:
- I like that each provider is forced to convert the account to an ID that uniquely identifies it. But I wonder how this would work for some providers such as the password provider. To be truly generalist, I think that the password provider should _not_ be specialized to any specific account identifier (e.g. e-mail, password…).
- How would providers write a `provider` string? Are components supposed to automatically infer it from the current component path?
- It looks to me that this design make it a little bit harder to enforce in the DB structure some invariants about the app. For example, in an app where users are always tied to a single Google account, it’s not possible to enforce in the DB schema that `accounts` has only a single row per user (other than writing a `createOrUpdateUser` function that does this)

Nicolapps (approved):
Thanks! I have some feedback about the design, but we can address it later to avoid blocking the other PRs.

Can you please have a look at this message before merging? https://convexdev.slack.com/archives/C04U4LQ9WDD/p1782692326691269?thread_ts=1782493093.814619&cid=C04U4LQ9WDD

examples/react-minimal/package.json
dowski: Yep!

src/components/core/crypto.ts
dowski: It's for claims from provider, and yes, since we don't have providers yet it's unused. `iss`, `exp`, `sub`, etc. are passed in and set directly below.

src/components/core/public.ts
dowski: If by switch to sessions you mean pull out JWTs and use plain session ID strings in their place, I'm not sure we're fully aligned on that. We might continue using the JWT but with a session ID bundled inside. We can discuss that further though.

src/components/core/public.ts
dowski: Yep!

src/components/core/public.ts
dowski: Yep good call.

src/components/core/public.ts
dowski: Thanks! I've fixed up these usages, and I'll add the ESLint plugin in a followup.

src/components/core/schema.ts
dowski: Yeah, this is fine to leave off for now I think. Thanks for the cleanup suggestion.

src/components/core/schema.ts
dowski: Thanks for the thoughtful critique here.

* We can discuss the password provider and options in your internal design for it.
* Providers define their own "provider" string.
* Yep, enforcing that invariant wasn't a goal here as it's intended to support multiple providers.

src/components/core/setup.ts
dowski: Good call.

src/components/core/setup.ts
dowski: I'm not sure that conforming to OIDC should be an explicit goal, but we can revisit that.

src/lib/types.ts
dowski: Great suggestions, I cleaned it up significantly based on your feedback.

Re: typesafe `ConvexError`, let's explore how we might do that.

src/lib/types.ts
dowski: Yeah, having this be typesafe would be nice for sure. We should explore more options here.

src/lib/types.ts
dowski: Good suggestion!

src/lib/types.ts
dowski: I don't think we can type it that way because this is inside of the component which doesn't have knowledge of the app's schema or types.

.eslintrc.cjs
dowski: I just left the prototype code in the subrepo ignored. Maybe we'll even drop that.

Nicolapps: Thank you! Yes, to clarify this suggestion was about `examples`/`scripts`, I think it’s fine for the prototype to be ignored

src/lib/types.ts
Nicolapps: I think that it would work for `CreateOrUpdateUserFn` to use `GenericId<"users">` (because this type is only used in app code and does not go through component codegen), and for `CreateOrUpdateUserFunctionHandle` to type the return type as `string`. What do you think?

src/components/core/crypto.ts
Nicolapps: What kind of claims would we want to have providers include?

I wonder if letting the provider set any claims might be too permissive. There are some meaningful claims that we don’t set (like `jti`/`nbf`). Also even for the keys that we override, it might be misleading

What do you think about throwing an error here if any of the standard keys of `JWTPayload` is set? (`iss`, `sub`, `aud`, `jti`, `nbf`, `exp`, `iat`)

src/components/core/public.ts
Nicolapps: Sounds good. My initial understanding about sessions vs. JWTs [incorrectly](https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html#session-expiration) [assumed](https://app.notion.com/p/convex-dev/Convex-Auth-v2-Auth-session-storage-JWTs-vs-session-strings-382b57ff32ab804593ccf5b024d97df3?source=copy_link) that we wouldn’t need for sessions to expire.

src/lib/types.ts
```diff
@@ -43,29 +43,40 @@ export const vAuthClaims = v.object({
...
+ * to integrate its user model with Convex Auth.
+ *
+ * It will be called in two scenarios, both associated with a user signing in:
+ *
+ *  1. The first time a user signs in with an account from a provider.
+ *    * The `userId` argument will not be present in this case.
+ *    * The application should create a new user record or use trusted data
+ *      (e.g. a verified email) in the `profile` to link to an existing
+ *      user record.
+ *    * The `_id` of the newly created user should be the return value.
```
Nicolapps: Thanks for updating the wording! It’s much clearer now.

Some additional feedback about the new wording:

> _“The `_id` of the newly created user should be the return value.”_

→ This phrasing might be confusing, because “newly created user” only applies in some cases. Also, “to link to an existing user record” is a little bit unclear under the current phrasing (it’s not super clear what “link” means here)

src/components/core/crypto.ts
dowski: Actually, I think we can just remove this. Thanks for pressing on it. Can resurrect if we wind up needing it.

src/lib/types.ts
dowski: Thanks, updated!

src/lib/types.ts
dowski: Ah nice, thanks. Seems like `GenericId<"users">` works in `CreateOrUpdateUserFunctionHandle` too.

PR #360, opened 2026-07-02
Password provider: Component functions

Nicolapps:
This implements the password provider component, as described in the following design doc: https://app.notion.com/p/convex-dev/Convex-Auth-v2-Passwords-38ab57ff32ab80a4b395f94378ebb958?source=copy_link

For now this only implements the component functions. A subsequent PR will add `setup*` functions that run in the app (root component), and contain the glue code for username/password necessary to use the password component in a real app.

This PR implements argon2id by using the `argon2` Rust crate. It’s maintained by the RustCrypto organization and [considered a “blessed” crate](https://blessed.rs/crates#section-cryptography). The Rust code is compiled to a WebAssembly file (71.1 KiB[^1]).

[^1]: The Wasm esbuild bundler in the Convex CLI encodes the contents of the Wasm file as base64, so the size of the actual module is slightly larger (99.6 KiB). If we wanted to reduce the module size, we could modify the push process to accept binary WASM files. I think the module size is fine for now, because this will only be loaded when the password component functions run, which is a small fraction of a real app workload.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
   "exports": {
     "./lib/*": "./src/lib/*",
     "./core/*": "./src/components/core/*",
+    "./passwordProvider/*": "./src/components/passwordProvider/*",
     "./server": {
       "types": "./dist/server/index.d.ts",
       "import": "./dist/server/index.js"
@@ -75,6 +76,7 @@
     }
   },
   "dependencies": {
+    "@convex-dev/rate-limiter": "^0.3.2",
     "@oslojs/crypto": "^1.0.1",
     "@oslojs/encoding": "^1.1.0",
     "cookie": "^1.0.1",
@@ -112,6 +114,7 @@
     "cspell": "^10.0.0",
     "dotenv": "^17.0.0",
     "eslint": "8.49.0",
+    "hash-wasm": "^4.12.0",
     "inquirer": "^14.0.0",
     "next": "^15.2.5",
     "npm-run-all2": "^9.0.0",
@@ -123,6 +126,7 @@
     "tsup": "^8.0.1",
     "typescript": "^6.0.3",
     "valibot": "^1.0.0",
+    "vite": "^8.0.16",
     "vitest": "^4.1.0"
   }
 }
--- a/src/components/passwordProvider/convex.config.ts
+++ b/src/components/passwordProvider/convex.config.ts
@@ -1,0 +1,17 @@
+import { defineComponent } from "convex/server";
+import rateLimiter from "@convex-dev/rate-limiter/convex.config.js";
+
+/**
+ * The password provider component.
+ *
+ * Securely stores and verifies passwords, keyed only by an opaque `userId`. It
+ * holds no email/username and knows nothing about the identity behind a user id
+ * — intentionally reusable in any context. The app owns the users table and
+ * maps its own identifiers onto the `userId` it passes in.
+ *
+ * Mounts the rate-limiter component to throttle `verifyPassword` per user id.
+ */
+const component = defineComponent("authPasswordProvider");
+component.use(rateLimiter);
+
+export default component;
--- a/src/components/passwordProvider/passwordProvider.test.ts
+++ b/src/components/passwordProvider/passwordProvider.test.ts
@@ -1,0 +1,277 @@
+import { convexTest } from "convex-test";
+import { describe, expect, test } from "vitest";
+import { register as registerRateLimiter } from "@convex-dev/rate-limiter/test";
+import {
+  argon2d,
+  argon2i,
+  argon2id,
+  argon2Verify,
+  bcrypt,
+  bcryptVerify,
+} from "hash-wasm";
+import { api } from "./_generated/api.js";
+import schema from "./schema.js";
+import { hashPassword, verifyPassword } from "./argon2.js";
+
+const modules = import.meta.glob("./**/*.ts");
+
+const PASSWORD = "correct horse battery staple"; // 28 chars, valid
+
+function setup() {
+  const t = convexTest(schema, modules);
+  // The component mounts the rate limiter; register it with the test instance
+  // so `verifyPassword`'s throttle has a backing component.
+  registerRateLimiter(t);
+  return t;
+}
+
+describe("setPassword + verifyPassword", () => {
+  test("verifies the correct password after setting it", async () => {
+    const t = setup();
+    await t.mutation(api.public.setPassword, {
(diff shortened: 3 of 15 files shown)


Nicolapps:
* **#371** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/371?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#368** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/368?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#360** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/360?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/360?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#365** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/365?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#362** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/362?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#364** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/364?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `reboot`


<h2></h2>

package.json
```diff
@@ -66,6 +67,7 @@
     }
   },
   "dependencies": {
+    "@convex-dev/rate-limiter": "^0.3.2",
```
Nicolapps, line 79: It’s annoying that we have to add this real dependency to the whole package. Presumably, before we publish any of this, we’ll evolve this repo to be a monorepo (e.g. pnpm workspace), so that we can publish providers independently?

This will be even more important when we’ll need to add peer dependency (e.g. on `@convex-dev/resend` for the email validation component).

src/components/passwordProvider/convex.config.ts
```diff
@@ -0,0 +1,17 @@
...
+ * The password provider component.
+ *
+ * Securely stores and verifies passwords, keyed only by an opaque `userId`. It
+ * holds no email/username and knows nothing about the identity behind a user id
+ * — intentionally reusable in any context. The app owns the users table and
+ * maps its own identifiers onto the `userId` it passes in.
+ *
+ * Mounts the rate-limiter component to throttle `verifyPassword` per user id.
+ */
+const component = defineComponent("authPasswordProvider");
```
Nicolapps, line 14: I’m not fully sure about the naming here.

- I think we’ll want to publish all auth packages as `@convex-dev/auth-*`. To this might be something like `@convex-dev/auth-password-provider`
- I’m not sure if we want to include `provider` in the login method component names. In one sense it can be helpful to distinguish provider components (e.g. password, passkeys, OAuth…) from non-provider components (e.g. `@convex-dev/auth-email-validation`). At the same time, I wonder if this naming convention makes sense when these components are used in other contexts (e.g. the GitHub example in Christian’s prototype)

We can always revisit this later if we want

package.json
```diff
@@ -99,10 +101,11 @@
     "@types/react": "^18.3.12",
     "@typescript-eslint/eslint-plugin": "^6.18.1",
     "chalk": "^5.3.0",
-    "convex-test": "^0.0.20",
+    "convex-test": "^0.0.53",
     "cspell": "^10.0.0",
     "dotenv": "^17.0.0",
     "eslint": "8.49.0",
+    "hash-wasm": "^4.12.0",
```
Nicolapps, line 117: Adding a dev dependency on `hash-wasm` for tests only (I wanted to have a reference implementation we can compare ours with)

src/components/passwordProvider/public.ts
```diff
@@ -0,0 +1,93 @@
...
+ */
+export const setPassword = mutation({
+  args: { userId: v.string(), password: v.string() },
+  returns: v.null(),
+  handler: async (ctx, { userId, password }) => {
+    assertValidPassword(password);
+    const passwordHash = await hashPassword(normalizePassword(password));
+    const existing = await passwordByUserId(ctx, userId);
+    if (existing !== null) {
+      await ctx.db.patch(existing._id, { passwordHash });
```
dowski: It's best to include the table name here, right?

src/components/passwordProvider/public.ts
```diff
@@ -0,0 +1,93 @@
...
+ * Set (or replace) a user's password.
+ *
+ * This can be used both when creating a new user (will create a row),
+ * or to update the password for an existing user (will update the existing row).
+ */
+export const setPassword = mutation({
+  args: { userId: v.string(), password: v.string() },
+  returns: v.null(),
+  handler: async (ctx, { userId, password }) => {
+    assertValidPassword(password);
```
dowski, line 46: Should there be any checks on `userId`? Like no empty string, no extremely long strings, etc?

package.json
dowski: Yeah we should come to a firm conclusion on this soon. I'm not sold on individual packages yet though.

src/components/passwordProvider/passwordProvider.test.ts
dowski, line 1: Great tests!

src/components/passwordProvider/schema.ts
```diff
@@ -0,0 +1,9 @@
+import { defineSchema, defineTable } from "convex/server";
+import { v } from "convex/values";
+
+export default defineSchema({
+  passwords: defineTable({
+    userId: v.string(),
+    passwordHash: v.string(), // PHC string
```
dowski: nit: consider including PHC in the name in some form (rather than a comment) since it's a specially encoded string.

dowski (approved):

src/components/passwordProvider/public.ts
Nicolapps: Yep, thanks. Normally I use the ESLint plugin which fixes this automatically for me, so I don’t pay attention 😛

src/components/passwordProvider/public.ts
Nicolapps: I prefer not to make any assumptions about this for now, because we will probably allow users to choose custom ID formats in the future

src/components/passwordProvider/schema.ts
Nicolapps: I don't have strong opinions about this, I renamed it in eb64fd6642d41d77fa97d74a2a2b94a5f6064fc4

PR #366, opened 2026-07-02
Updated core API with better type safety

dowski:
The `setupCore` method now takes a list of [ProviderConfig, Options] tuples
and ensures that the application's `createOrUpdateUsers` function is typed
so its `provider` argument matches the configured providers[^1].

This necessitated splitting out an `attachUserCallback` function which is
returned by `setupCore`. Without that, there are circular typing issues
since TypeScript is trying to determine the type of `createOrUpdateUser`
which is a part of the application's API while also exporting the
other functions that are returned during setup (that also become part of
the application's API).

Another benefit of this change is that providers get a more specificly
typed "shape" which should help guide proper creation and configuration.

I took a pass over the documentation too and expanded/moved some
to where it would be more accessible. If you check out the PR and
look at the `examples/react-minimal/auth.ts` you should see helpful
docs on most of the key symbols that are used when creating and
configuring providers. 

You can see an example of usage of the API here: https://github.com/get-convex/convex-auth/pull/367

----

[^1]: The matching is a little backward compared to regular TS function references though - it will catch a `createOrUpdateUser` function that supports too many providers, not too few.

### DIFF
--- a/examples/react-minimal/.gitignore
+++ b/examples/react-minimal/.gitignore
@@ -1,0 +1,2 @@
+
+.env.local
--- a/examples/react-minimal/convex/_generated/api.d.ts
+++ b/examples/react-minimal/convex/_generated/api.d.ts
@@ -1,0 +1,53 @@
+/* eslint-disable */
+/**
+ * Generated `api` utility.
+ *
+ * THIS CODE IS AUTOMATICALLY GENERATED.
+ *
+ * To regenerate, run `npx convex dev`.
+ * @module
+ */
+
+import type * as auth from "../auth.js";
+import type * as users from "../users.js";
+
+import type {
+  ApiFromModules,
+  FilterApi,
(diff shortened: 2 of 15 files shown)


erquhart (approved):
Overall approach makes sense, fixes the covariant/contravariant problem without complicating the interface 👍

Assuming the type issues I'm seeing locally and a vitest env issue in this PR's linting are known and accounted for.

dowski:
> Overall approach makes sense, fixes the covariant/contravariant problem without complicating the interface 👍

Actually that isn't fixed here - need some change to `FunctionReference` (as discussed elsewhere) to make that happen.

> Assuming the type issues I'm seeing locally and a vitest env issue in this PR's linting are known and accounted for.

Yeah, those will be fixed in the follow up, thanks for flaggging.

PR #367, opened 2026-07-02
Anonymous provider

dowski:
This is an anonymous provider for Convex Auth v2. See `examples/react-minimal/convex/auth.ts` for usage. It allows signing in as an anonymous user. The account ID is a random UUID stored in the `anonymous` component's `accounts` table.

> [!NOTE]
> This could potentially be simpler than a component, but it will at least serve as a complete example of what a minimal auth provider component looks like.

Includes unit tests and some integration-style tests with the `core` component (`examples/react-minimal/convex/auth.test.ts`).

----


packages/core/src/components/anonymous/convex.config.ts
```diff
@@ -0,0 +1,5 @@
+import { defineComponent } from "convex/server";
+
+const component = defineComponent("anonymous");
```
Nicolapps:
```suggestion
const component = defineComponent("authAnonymous");
```

This name is going to be the default path for component installation if the user doesn’t change it. I think we should prefix all the auth components with `auth*` to make it clearer to the user.

I also think it’s fine to not include “provider” in our component names. I think it’s already clear enough, and we can always group all providers together in the docs.

convex.json
```diff
@@ -0,0 +1,4 @@
+{
```
Nicolapps, line 1: We might remove this when we have more demos to remove confusion, but it’s probably fine for now

packages/core/src/components/anonymous/setup.ts
```diff
@@ -0,0 +1,53 @@
...
+ * The {@link AnonymousOptions.allowReturningAccounts} value controls whether a
+ * previously issued anonymous ID can be used to re-establish a session. It
+ * defaults to `false`.
+ */
+export const Anonymous = defineProvider({
+  name: "anonymous",
+  setup: (completeSignIn, options: AnonymousOptions) => {
+    const { component, allowReturningAccounts = false } = options;
+    return {
+      signInAnonymous: actionGeneric({
```
Nicolapps:
```suggestion
      signInAnonymous: mutationGeneric({
```

packages/core/src/components/anonymous/setup.ts
```diff
@@ -0,0 +1,53 @@
...
+            providerAccountId: anonymousId,
+            profile: {},
+          });
+        },
+      }),
+    };
+  },
+});
+
+type AnonymousOptions = {
```
Nicolapps: Suggestion: it would make sense to export this type, since we refer to it in the documentation

packages/core/src/components/anonymous/provider.ts
```diff
@@ -0,0 +1,39 @@
...
+export const signInAnonymous = mutation({
+  args: { id: v.optional(v.string()) },
+  handler: async (ctx, args) => {
+    const returningId = args.id ?? null;
+    if (returningId === null) {
+      const anonymousId = crypto.randomUUID();
+      await ctx.db.insert("accounts", { lastSignIn: Date.now(), anonymousId });
+      return anonymousId;
+    }
+    const account = await getAccountByAnonymousId(ctx, returningId);
```
Nicolapps: Shouldn’t this use lastSignIn?

examples/react-minimal/convex/auth.ts
```diff
@@ -1,38 +1,21 @@
...
 // The core owns sessions, accounts, and JWT minting. It calls back into our
 // `upsertFromAuth` on every sign-in from a provider.
-export const { signOut, refreshSession, providers } = setupCore({
+export const {
+  signOut,
+  refreshSession,
+  providers: {
+    anonymous: { signInAnonymous },
+  },
+} = setupCore({
```
Nicolapps, line 13: I double checked, and this is indeed equivalent to

```ts
const core = setupCore({ … });
export const signOut = core.signOut;
export const refreshSession = core.refreshSession;
export const signInAnonymous = core.providers.anonymous.signInAnonymous;
```

without `providers` or `anonymous` being exported.

I hope this syntax won’t be too confusing to users.

examples/react-minimal/convex/schema.ts
```diff
@@ -11,5 +11,7 @@ export default defineSchema({
   }).index("by_author", ["authorId"]),
   users: defineTable({
     name: v.string(),
+    /** Used to record an anonymous account ID for this user, for signing back in. */
+    anonymousAccountId: v.optional(v.string()),
```
Nicolapps: Why do we record the anonymous account ID here too?

I’m a little worried about having this in the app tables, since:
- it effectively works as a secret (an app might leak the entire `users` row by mistake)
- someone could unintentionally change the anonymous account ID in the app but not in the component

Maybe an alternative would be to store nothing in `users`, and add a component query to get the current user ID?

examples/react-minimal/convex/users.ts
```diff
@@ -8,17 +9,41 @@ import { v } from "convex/values";
...
     userId: v.union(v.string(), v.null()),
   },
   returns: v.id("users"),
   handler: async (ctx, args) => {
-    if (args.profile.name && typeof args.profile.name === "string") {
-      return ctx.db.insert("users", { name: args.profile.name });
+    if (args.userId) {
+      // We don't do any updating for existing users. Simply return
+      // the ID back.
+      return args.userId as Id<"users">;
```
Nicolapps: Shouldn’t we type upsertFromAuth in a way so that `userId` is a `v.id("users")` (in app code)?

packages/core/src/components/anonymous/provider.ts
Nicolapps: (In general, I’m not sure I understand why it would be useful for the anonymous component specifically to store a `lastSignIn` field. Maybe it would work better to remove the field entirely for now.)

packages/core/src/components/anonymous/schema.ts
```diff
@@ -0,0 +1,9 @@
+import { defineSchema, defineTable } from "convex/server";
+import { v } from "convex/values";
+
+export default defineSchema({
+  accounts: defineTable({
```
Nicolapps, line 5: > This could potentially be simpler than a component, but it will at least serve as a complete example of what a minimal auth provider component looks like.

I agree, maybe we could consider changing this to be only helper functions before releasing v2? It would be worth having a TODO comment about this

Nicolapps (approved):

examples/react-minimal/convex/schema.ts
dowski: Stripped this out.

examples/react-minimal/convex/users.ts
dowski: Removed this (but yes if we kept it).

packages/core/src/components/anonymous/convex.config.ts
dowski: Done.

packages/core/src/components/anonymous/provider.ts
dowski: Removed.

packages/core/src/components/anonymous/setup.ts
dowski: Done.

packages/core/src/components/anonymous/setup.ts
dowski: Dropped it.

convex.json
dowski: Sounds good.

packages/core/src/components/anonymous/provider.ts
Nicolapps: We are not allowing users to log in again with an anonymous account, is that right?

dowski: Yep that's right.

PR #364, opened 2026-07-02
Fix Auth v2 CI

Nicolapps:
- Making sure the v2 CI also runs when pushing stacked PRs
- Fix the tests that are run: we want to run the Vitest tests, not `test:once` (which runs the v1 tests in `/test`)

PR #362, opened 2026-07-02
Password provider: Rust crate for argon2id hashes

Nicolapps:
This PR is the first part of the password component described at https://app.notion.com/p/convex-dev/Convex-Auth-v2-Passwords-38ab57ff32ab80a4b395f94378ebb958?source=copy_link. As described in the design doc, we’re going to implement Argon2id hashing using a small WebAssembly module we’re going to embed in the component. This PR adds this module, while the actual component itself will be implemented in subsequent PRs.

This module is a thin wrapper around the `argon2` Rust crate. It’s maintained by the RustCrypto organization and [considered a “blessed” crate](https://blessed.rs/crates#section-cryptography). The Rust code is compiled to a WebAssembly file (66 KiB[^1]).

[^1]: The Wasm esbuild bundler in the Convex CLI encodes the contents of the Wasm file as base64, so the size of the actual module is slightly larger. If we wanted to reduce the module size, we could modify the push process to accept binary WASM files. I think the module size is fine for now, because this will only be loaded when the password component functions run, which is a small fraction of a real app workload.

PR #368, opened 2026-07-06
Password provider: Improve error return values

Nicolapps:
This changes how the password provider returns errors. The goal is to:
- Make sure the expected user errors (e.g. password is too long) does _not_ throw an error, in order to not clutter the app logs
- For errors that are unrecoverable (e.g. unexpected issue in WASM), throw a clear error that will appear in the user’s logs, and don’t return a value.
- Make user errors clear at the type level. This allows errors to be forwarded to the client. Doing this instead of hardcoding error messages in `ConvexError`s will make it easier for users to localize error messages, and customize to the application context.
- Make sure the ergonomics of error handling in `setup*` functions (especially if they’re customized by users) are good. With the return value format that I chose, it should be pretty straightforward:

```ts
const result = await ctx.runMutation(
  components.authPassword.validatePassword,
  { userId, password },
);

if (!result.success) {
  return result.userError; // e.g. { error: "PASSWORD_TOO_LONG", maximumLength: 100 }
}

// …
```

### DIFF
--- a/src/components/passwordProvider/public.ts
+++ b/src/components/passwordProvider/public.ts
@@ -1,13 +1,17 @@
 import { mutation, QueryCtx } from "./_generated/server";
 import { Doc } from "./_generated/dataModel";
 import { components } from "./_generated/api";
-import { v } from "convex/values";
+import { Infer, v } from "convex/values";
 import { RateLimiter, MINUTE } from "@convex-dev/rate-limiter";
 import {
   hashPassword,
   verifyPassword as verifyPasswordHash,
 } from "./argon2.js";
-import { assertValidPassword, normalizePassword } from "./validation.js";
+import {
+  validatePassword,
+  normalizePassword,
+  setPasswordUserError,
+} from "./validation.js";
 
 // Throttle for password verification attempts.
 //
@@ -33,6 +37,12 @@ const rateLimiter = new RateLimiter(components.rateLimiter, {
   },
 });
 
+const setPasswordResult = v.union(
+  v.object({ success: v.literal(true) }),
+  v.object({ success: v.literal(false), userError: setPasswordUserError }),
+);
+type SetPasswordResult = Infer<typeof setPasswordResult>;
+
 /**
  * Set (or replace) a user's password.
  *
@@ -41,9 +51,13 @@ const rateLimiter = new RateLimiter(components.rateLimiter, {
  */
 export const setPassword = mutation({
   args: { userId: v.string(), password: v.string() },
-  returns: v.null(),
-  handler: async (ctx, { userId, password }) => {
-    assertValidPassword(password);
+  returns: setPasswordResult,
+  handler: async (ctx, { userId, password }): Promise<SetPasswordResult> => {
+    const userError = validatePassword(password);
+    if (userError !== null) {
+      return { success: false, userError };
+    }
+
     const passwordHashPHC = await hashPassword(normalizePassword(password));
     const existing = await passwordByUserId(ctx, userId);
     if (existing !== null) {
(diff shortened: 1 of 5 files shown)


Nicolapps:
* **#371** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/371?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#368** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/368?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/368?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#360** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/360?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#365** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/365?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#362** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/362?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#364** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/364?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `reboot`


<h2></h2>

src/components/passwordProvider/public.ts
```diff
@@ -41,20 +51,36 @@ const rateLimiter = new RateLimiter(components.rateLimiter, {
...
   },
 });
 
+const verifyPasswordUserError = v.union(
+  v.object({ error: v.literal("INVALID_CREDENTIALS") }),
+  v.object({ error: v.literal("RATE_LIMITED"), retryAfterMs: v.number() }),
+);
+
+const verifyPasswordResult = v.union(
+  v.object({ success: v.literal(true) }),
```
Nicolapps, line 88: This will need to include some kind of object the user will pass to the core component to create a session

src/components/passwordProvider/public.ts
```diff
@@ -41,20 +51,36 @@ const rateLimiter = new RateLimiter(components.rateLimiter, {
  */
 export const setPassword = mutation({
   args: { userId: v.string(), password: v.string() },
-  returns: v.null(),
-  handler: async (ctx, { userId, password }) => {
-    assertValidPassword(password);
+  returns: setPasswordResult,
+  handler: async (ctx, { userId, password }): Promise<SetPasswordResult> => {
+    const userError = validatePassword(password);
```
dowski: nit: maybe call the function `validateNewPassword`. I got a little crossed up between validate/verify.

src/components/passwordProvider/public.ts
dowski: The model we're working with right now has the provider returning an object that the core will use to initiate a session:

https://github.com/get-convex/convex-auth/blob/cf62164952e0abdb79a292df351d5fb81b5af6da/src/lib/types.ts#L27-L51

Nicolapps: Yep, this is what I have in mind

src/components/passwordProvider/public.ts
Nicolapps: Thanks, I agree this was confusing. I renamed the function to `validatePasswordInputFormat` in bfdb82b. I also made sure `verifyPassword` also performs the same input validation in ccd54f7 (I added a comment explaining my motivation here)

src/components/passwordProvider/public.ts
dowski: Cool - and really, that can happen in the `setup*` level code so this is probably fine here.

Nicolapps: Yes. Ideally we have a way to make sure the object that we return can only be used safely, maybe with a branded type, for cases where users manually write setup functions.

dowski (approved):

Nicolapps:
Thanks ! I'm merging this PR now. There are lint errors but they seem to come from the v1 code. These are going to go away when #370 is merged.

PR #369, opened 2026-07-07
Updated core API with better type safety

dowski:
*Originally merged in #366, reverted in d4358e88db3ed992c5d6080ec861efba1cfb11cc and brought back here for more discussion ahead of merging.*

The `setupCore` method now takes a list of [ProviderConfig, Options] tuples
and ensures that the application's `createOrUpdateUsers` function is typed
so its `provider` argument matches the configured providers.

This necessitated splitting out an `attachUserCallback` function which is
returned by `setupCore`. Without that, there are circular typing issues
since TypeScript is trying to determine the type of `createOrUpdateUser`
which is a part of the application's API while also exporting the
other functions that are returned during setup (that also become part of
the application's API).

Another benefit of this change is that providers get a more specificly
typed "shape" which should help guide proper creation and configuration.

----

### DIFF
--- a/examples/react-minimal/convex/auth.ts
+++ b/examples/react-minimal/convex/auth.ts
@@ -1,13 +1,38 @@
+import { defineProvider } from "@convex-dev/auth/lib/types.js";
 import { components, internal } from "./_generated/api";
-import { setupCore } from "@convex-dev/auth/core/setup.js";
+import { provider, setupCore } from "@convex-dev/auth/core/setup.js";
 
-// The core owns sessions, accounts, and JWT minting. It calls back into our
-// `upsertFromAuth` on every sign-in; this example owns no users table and just
-// echoes the account id back as the app user id.
-const core = setupCore({
-  component: components.core,
-  createOrUpdateUser: internal.users.upsertFromAuth,
+/**
+ * Documentation for the FakeProvider.
+ *
+ * Pair with {@link FakeOptions} when installing as a {@link provider}.
+ */
+// TODO: dowski - remove this when we have real providers to work with
+const FakeProvider = defineProvider({
+  name: "fake",
+  setup: (completeSignIn, options: FakeOptions) => {
+    return {
+      login: (arg1: string) => null,
+    };
+  },
 });
 
-// A sign-in path (e.g. username/password) is wired in alongside a provider.
-export const { signOut, refreshSession } = core;
+/**
+ * Docs for the FakeOptions.
+ */
+type FakeOptions = {
+  /**
+   * Docs for the exampleValue.
+   */
+  exampleValue: string;
+};
+
+// The core owns sessions, accounts, and JWT minting. It calls back into our
+// `upsertFromAuth` on every sign-in from a provider.
+export const { signOut, refreshSession, providers } = setupCore({
+  component: components.core,
+  providers: [
+    // TODO: dowski - remove this when we have real providers to work with
+    provider(FakeProvider, { exampleValue: "passed to FakeProvider.setup" }),
+  ],
+}).attachUserCallback(internal.users.upsertFromAuth);
--- a/examples/react-minimal/convex/users.ts
+++ b/examples/react-minimal/convex/users.ts
@@ -1,39 +1,24 @@
-import { internalMutation, query } from "./_generated/server";
+import { internalMutation } from "./_generated/server";
 import { v } from "convex/values";
 
 /**
  * The app's user create-or-update callback. The core invokes it (via a function
  * handle) on every sign-in — without a `userId` the first time an identity is
  * seen, and with the resolved `userId` thereafter.
- *
- * This minimal example owns no users table. Instead of storing a row, it simply
- * echoes the provider-scoped account id back as the app's user id. The core
- * treats that value as an opaque string: it goes into the `accounts`/`sessions`
- * tables and becomes the JWT `subject`, so `ctx.auth.getUserIdentity().subject`
- * ends up being the provider account id. The (provider, providerAccountId) pair
- * is stable, so the subject stays the same across logins.
- *
- * The `userId` arg is part of the core's callback contract (it's set on a
- * returning sign-in, and when an existing user links another identity). With no
- * users table there's nothing to update, so we just honor it when present.
  */
 export const upsertFromAuth = internalMutation({
   args: {
-    provider: v.string(),
+    // TODO: dowski - remove this when we have real providers to work with
+    provider: v.union(v.literal("fake")),
     providerAccountId: v.string(),
     profile: v.any(),
-    userId: v.optional(v.string()),
+    userId: v.union(v.string(), v.null()),
   },
-  returns: v.string(),
-  handler: async (_ctx, args) => args.userId ?? args.providerAccountId,
-});
-
-/**
- * The signed-in identity, straight from the access token. There's no user row
- * to load — the subject is the provider account id the core minted into the
- * JWT. Returns null when no one is signed in.
- */
-export const loggedInUser = query({
-  args: {},
-  handler: async (ctx) => ctx.auth.getUserIdentity(),
+  returns: v.id("users"),
+  handler: async (ctx, args) => {
+    if (args.profile.name && typeof args.profile.name === "string") {
+      return ctx.db.insert("users", { name: args.profile.name });
+    }
+    throw Error("unable to add user; no name");
+  },
 });
--- a/packages/core/src/components/core/setup.ts
+++ b/packages/core/src/components/core/setup.ts
@@ -1,29 +1,147 @@
 import {
   mutationGeneric,
   createFunctionHandle,
-  type GenericActionCtx,
-  type GenericDataModel,
+  RegisteredMutation,
 } from "convex/server";
 import { v } from "convex/values";
 import type { ComponentApi } from "./_generated/component";
 import {
   vTokenBundle,
   type TokenBundle,
-  type AuthClaims,
-} from "../../lib/types";
-import { CreateOrUpdateUserFn } from "../../lib/types";
+  ProviderConfig,
+  CompleteSignInFunc,
+  CreateOrUpdateUserFn,
+} from "../../lib/types.js";
+
+// Helper: pairs a `ProviderConfig` with its options, preserving `Name`, `Options` and `Api` individually.
+// TODO: dowski - consider whether this function should exist, or whether defineProvider
+// should return a function that's callable with the options.
+export function provider<Name extends string, Options, Api>(
+  config: ProviderConfig<Name, Options, Api>,
+  options: Options,
+): readonly [ProviderConfig<Name, Options, Api>, Options] {
+  return [config, options] as const;
+}
+
+// An explicitly loosley typed tuple of [ProviderConfig, Options].
+type ProviderWithOptions = readonly [ProviderConfig<any, any, any>, any];
+
+// Maps a tuple of entries to
+// { [ProviderConfig.name]: ReturnType<ProviderConfig.setup> } by distributing
+// over the tuple's union.
+type ProviderApis<T extends readonly ProviderWithOptions[]> = {
+  [K in T[number]as K[0]["name"]]: ReturnType<K[0]["setup"]>;
+};
+
+/**
+ * Refreshes a session using the given token, rotating the refresh token.
+ *
+ * A successful refresh returns a new token bundle. A failed refresh (an
+ * unknown or expired token) returns `null` and leaves no usable session
+ * behind (an expired session is deleted as part of the same call). Callers,
+ * including the client, should treat a `null` result as signed-out and clear
+ * any stored session.
+ */
+type RefreshSessionFunc = RegisteredMutation<
+  "public",
+  {
+    refreshToken: string;
+  },
+  Promise<TokenBundle | null>
+>;
+
+/**
+ * Signs out of the current session.
+ *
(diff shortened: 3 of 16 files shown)


examples/react-minimal/convex/auth.ts
```diff
@@ -1,13 +1,33 @@
...
+        },
+      },
+      { x: 2 },
+    ),
+  ],
+}).attachUserCallback(internal.users.upsertFromAuth);
 
-// A sign-in path (e.g. username/password) is wired in alongside a provider.
-export const { signOut, refreshSession } = core;
+const { foo, bar } = providers;
```
Nicolapps: What is this supposed to do?

If this is meant to be `export const { foo, bar } = providers`, and the plan is to export provider-defined functions, this doesn’t seem to work at the moment with Convex.

For example, when pushing this code:

```ts
import { query, mutation } from "./_generated/server";

const myQuery = query({
  args: {},
  handler: async () => {
    return 42;
  },
});

const myMutation = mutation({
  args: {},
  handler: async () => {},
});

export const test = { myQuery, myMutation };
```

The Convex analyze process doesn’t find any functions:

Image

packages/core/src/lib/types.ts
```diff
@@ -50,34 +50,87 @@ export const vCreateOrUpdateUser = v.object({
   userId: v.union(v.string(), v.null()),
 });
 
+export type CreateOrUpdateUserFn<Provider extends string> = FunctionReference<
+  "mutation",
+  "internal",
+  {
+    provider: Provider;
+    providerAccountId: string;
+    profile: any;
```
Nicolapps, line 60: Do you plan to modify `CreateOrUpdateUserFn` so that `profile` is typesafe too?

examples/react-minimal/convex/auth.ts
```diff
@@ -1,13 +1,33 @@
+import { CompleteSignInFunc, defineProvider } from "../../../src/lib/types";
 import { components, internal } from "./_generated/api";
-import { setupCore } from "@convex-dev/auth/core/setup.js";
+import { provider, setupCore } from "@convex-dev/auth/core/setup.js";
+
+const FooProvider = defineProvider({
```
Nicolapps: Where do we expect for `defineProvider` to be called in practice? Presumably from a provider-provided `setup*` function, is that right?

I wonder if it would make sense to have the `name` be automatically inferred from the component path, for cases where the same provider component is installed multiple times (e.g. OAuth in case we decide that the OAuth component supports a single provider for each instantiation). Do you have any thoughts about this?

src/lib/types.ts
```diff
@@ -50,34 +50,87 @@ export const vCreateOrUpdateUser = v.object({
...
+   *
+   * This value will be persisted the `accounts` table and passed to
+   * the {@link CreateOrUpdateUserFn}.
+   */
+  name: N;
+  /**
+   * Convex Auth will call this function to setup this provider.
+   *
+   * It should return an object of names to `actionGeneric` Convex
+   * functions that will make up the public API surface of the provider.
```
Nicolapps: nit: it probably make sense to clarify here what “public” means (i.e. user-callable), because we also talk about “public” component functions for component functions that are callable by the app

packages/core/src/lib/types.ts
```diff
@@ -50,34 +50,87 @@ export const vCreateOrUpdateUser = v.object({
...
+
+export type CompleteSignInFunc = (
+  ctx: GenericActionCtx<any>,
+  claims: AuthClaims,
+) => Promise<TokenBundle>;
+
+type ProviderSetupFunc<O, P> = (
+  completeSignIn: CompleteSignInFunc,
+  options: O,
+) => P;
```
Nicolapps, line 74: Why are there no type restrictions on `P`?

With the current types, a provider setup function can return arbitrary values, and the arguments the actions take are not typed.

examples/react-minimal/convex/auth.ts
```diff
@@ -1,13 +1,33 @@
...
   component: components.core,
-  createOrUpdateUser: internal.users.upsertFromAuth,
-});
+  providers: [
+    provider(FooProvider, { x: "value" }),
+    provider(
+      {
+        name: "bar",
+        setup: (completeSignIn: CompleteSignInFunc, options: { x: number }) => {
+          return { x: "hi" };
```
Nicolapps: Isn’t `"hi"` here supposed to be an action handler instead of a string?

packages/core/src/components/core/setup.ts
```diff
@@ -1,28 +1,142 @@
...
 } from "../../lib/types.js";
 import { CreateOrUpdateUserFn } from "../../lib/types.js";
 
+// Helper: pairs a `ProviderConfig` with its options, preserving `Name`, `Options` and `Api` individually.
+export function provider<Name extends string, Options, Api>(
+  config: ProviderConfig<Name, Options, Api>,
+  options: Options,
+): readonly [ProviderConfig<Name, Options, Api>, Options] {
+  return [config, options] as const;
+}
```
Nicolapps, line 24: The generic typing of `provider` seems to cause a worst developer experience with editors:

- Users can’t ⌘-click a parameter name to go to the source
- Users can’t hover an parameter name to see the documentation of a parameter
- TypeScript seems to be more lenient than what necessary with the types. For example, it doesn’t complain when providing parameters that are wrong (e.g. `provider(FooProvider, { x: "value", y: "unexpected arg" })` in your example)

Maybe a potential solution would be to make `defineProvider` return a function? And then the provider instantiation would look like `FooProvider({ x: "value" })`

packages/core/src/components/core/setup.ts
```diff
@@ -1,28 +1,142 @@
...
+   *    * The existing `userId` must be the return value.
+   *
+   * The application keeps ownership of its users table — the core only holds a
+   * reference to this one mutation.
+   *
+   * If an application wants to reject a sign in, it can throw a `ConvexError`
+   * and the entire sign in attempt will be blocked.
+   */
+  attachUserCallback(
+    createOrUpdateUser: CreateOrUpdateUserFn<T[number][0]["name"]>,
```
Nicolapps, line 129: Where does the `T[number][0]["name"]` come from? Is that part of the type issues that still need to be fixed?

(I also have the same question for `buildFullApi`)

examples/react-minimal/convex/auth.ts
```diff
@@ -1,13 +1,33 @@
+import { CompleteSignInFunc, defineProvider } from "../../../src/lib/types";
 import { components, internal } from "./_generated/api";
-import { setupCore } from "@convex-dev/auth/core/setup.js";
+import { provider, setupCore } from "@convex-dev/auth/core/setup.js";
+
+const FooProvider = defineProvider({
+  name: "foo",
+  setup: (completeSignIn, options: { x: string }) => {
+    return {
+      login: () => null,
```
Nicolapps: The documentation of `ProviderConfig` says:

> It should return an object of names to `actionGeneric` Convex functions 

Is this supposed to be something like `login: action({ args: { … }, handler: async () => null})`?

src/lib/types.ts
```diff
@@ -50,34 +50,87 @@ export const vCreateOrUpdateUser = v.object({
...
+   * The name of this provider.
+   *
+   * This value will be persisted the `accounts` table and passed to
+   * the {@link CreateOrUpdateUserFn}.
+   */
+  name: N;
+  /**
+   * Convex Auth will call this function to setup this provider.
+   *
+   * It should return an object of names to `actionGeneric` Convex
```
Nicolapps: Why can providers only expose actions? Is there an advantage this restriction brings?

If we force providers to only add actions:
- It adds some overhead when the provider only needs to do mutations
- The app can’t compose the provider mutation with something that needs to run in the same transaction (and AFAIK we want for these functions to be composable, like in your challenges prototype)

examples/react-minimal/convex/users.ts
```diff
@@ -19,13 +19,19 @@ import { v } from "convex/values";
...
-    provider: v.string(),
+    provider: v.union(v.literal("foo"), v.literal("bar")),
     providerAccountId: v.string(),
     profile: v.any(),
-    userId: v.optional(v.string()),
+    userId: v.union(v.string(), v.null()),
+  },
+  returns: v.id("users"),
+  handler: async (ctx, args) => {
+    args.provider;
```
Nicolapps: Is it intentional to have this here? Or is it there to demonstrate the types?

packages/core/src/components/core/setup.ts
```diff
@@ -49,98 +190,66 @@ export function setupCore(opts: {
...
   refreshTokenTtlSeconds?: number;
-}) {
-  const {
-    component,
-    createOrUpdateUser,
-    accessTokenTtlSeconds,
-    refreshTokenTtlSeconds,
-  } = opts;
-
+  providers: T;
```
Nicolapps, line 210: Can you please document that field?

The `providers` array adds some level of abstraction that I think would benefit from some clarifications. From a user point of view, I think that I would be wondering:
- What does adding a provider in `providers` do? Does it automatically expose additional functions, which might allow users to do specific actions in the app?
- What are the consequences of removing a provider that was previously in the list? Does it invalidate sessions coming from that provider?

examples/react-minimal/convex/auth.ts
dowski: Yeah, `defineProvider` will be called by a provider component author. This is here because we don't have any providers yet. Generally the `Provider` gives a more definite shape to the previous `setup*` functions. Here are a couple examples from the anonymous provider.

1. [definition of the provider](https://github.com/get-convex/convex-auth/blob/238a829f0380d7c61306f6f98711eae9cfb2a26e/src/components/anonymous/setup.ts)
2. [hooking up a provider](https://github.com/get-convex/convex-auth/blob/238a829f0380d7c61306f6f98711eae9cfb2a26e/examples/react-minimal/convex/auth.ts)

I think using the component name is a good call. In fact in the anonymous provider I want to have two instances to test the different config options.

examples/react-minimal/convex/auth.ts
dowski: Generally yes: this was a stub implementation that exists just to prove out the types during dev. Technically the `setup` function can return anything, but it *should* return function that are meant to be exposed as a public Convex API.

examples/react-minimal/convex/auth.ts
dowski: Yep, removed this bad example.

examples/react-minimal/convex/auth.ts
dowski: Sorry for the bad/distracting code. This was simply used so I could look at the types that were inferred on the returned values. It's removed in the anonymous provider PR, but I'll remove it here too.

examples/react-minimal/convex/users.ts
dowski: This was another case of me dropping in a value to inspect its type. Cleaned up.

packages/core/src/components/core/setup.ts
dowski: I think this is worth exploring. Added a TODO.

packages/core/src/components/core/setup.ts
dowski: This is what builds the union of known providers that the `CreateOrUpdateUser` function should conform to. As discussed elsewhere though, the underlying `FunctionReference` treats the args covariantly.

packages/core/src/components/core/setup.ts
dowski: Done.

packages/core/src/lib/types.ts
dowski: I think that should be possible - added a TODO.

packages/core/src/lib/types.ts
dowski: This is no different than the previous `setup*` functions technically being able to return whatever they want. We can look into constraining it further if it is an issue though.

I'm not sure what you mean about the arguments not being typed though - the returned value should be fully typed. See `login` and `arg1` below.

Image

src/lib/types.ts
dowski: Yeah, fair point. I'll update this doc.

src/lib/types.ts
dowski: Clarified, thanks.

examples/react-minimal/convex/auth.ts
Nicolapps: Sounds good. How do we plan for provider functions to be exposed (so that app users can call them)?

packages/core/src/components/core/setup.ts
Nicolapps: Thanks!

packages/core/src/lib/types.ts
Nicolapps: We discussed this synchronously: the component is expected to provide public actions/mutations, and we probably want to make the types more strict here.

Nicolapps (approved):
Thanks for addressing all the feedback!

examples/react-minimal/convex/auth.ts
dowski: Also chatted about this synchronously. For posterity, they'll be exported either directly (via unpacking `providers`) or as separate `export const` statements.

dowski: And ideally, we can have them folded into the top level result and exported as peers to `signOut` and `refreshSession`.

PR #370, opened 2026-07-07
Remove Convex Auth v1 code + use pnpm

Nicolapps:
- **Remove Convex Auth v1 code**, in order to make it clearer what is the newer implementation and what is the old one
- **Remove the `prototype` submodule**. We can link manually to a checkout of the prototype in prompts when it’s useful.
- **Remove most agent skills**, because they aren't useful for what we're doing in particular. (I kept the create-component skill in `.agents` because this one can be useful. When using Claude, make sure to `@` this skill manually when necessary)
- **Use pnpm instead of NPM**. This will make it easier to work with multiple packages (in particular demos). Note that this PR does _not_ split providers to individual packages
- **Use ESLint 10 + Prettier**. I set up ESLint so that it uses recommended settings for TypeScript, including type-aware rules. I did not include ESLint rules for React because eslint-plugin-react does not support ESLint 10 yet. 
- **Adapt tooling** (e.g. GitHub actions) to match to the new structure. Note that I removed `‎.github/workflows/publish-every-commit.yml‎` for now in order, but we can restore it later when we know what the package structure is going to look like.


Nicolapps:
* **#370** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/370?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/370?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `reboot`


<h2></h2>

dowski (approved):

dowski:
Thanks!

Nicolapps:
image
Fixing this issue in a separate PR very soon to keep the diff simpler:

- #372

PR #387, opened 2026-07-09
Drop named types used internally for AuthApi

dowski:
Using them caused consumers to lose visibility into proper type of the actually returned functions. This is simpler and makes it easier to use correctly.

----


Nicolapps (approved):
Sounds good!

PR #388, opened 2026-07-09
Framework agnostic client APIs

dowski:
Framework agnostic primitives for interacting with Convex Auth.

Includes token storage and concurrent refresh support (via a mutex) and an API for exposing refresh and logout and communicating auth state changes out to observers (e.g. a React lib).

A React client building on these primitives is in #397.

----

### DIFF
--- a/packages/core/src/browser/mutex.ts
+++ b/packages/core/src/browser/mutex.ts
@@ -32,13 +32,10 @@ type MutexState = {
 
 const mutexes = new Map<string, MutexState>();
 
-function getMutex(key: string): MutexState {
-  let mutex = mutexes.get(key);
-  if (mutex === undefined) {
-    mutex = { currentlyRunning: null, waiting: [] };
-    mutexes.set(key, mutex);
-  }
-  return mutex;
+function manualMutex<T>(key: string, callback: () => Promise<T>): Promise<T> {
+  return new Promise<T>((resolve, reject) => {
+    enqueue(key, () => callback().then(resolve, reject));
+  });
 }
 
 function enqueue(key: string, callback: () => Promise<void>) {
@@ -56,8 +53,11 @@ function enqueue(key: string, callback: () => Promise<void>) {
   }
 }
 
-function manualMutex<T>(key: string, callback: () => Promise<T>): Promise<T> {
-  return new Promise<T>((resolve, reject) => {
-    enqueue(key, () => callback().then(resolve, reject));
-  });
+function getMutex(key: string): MutexState {
+  let mutex = mutexes.get(key);
+  if (mutex === undefined) {
+    mutex = { currentlyRunning: null, waiting: [] };
+    mutexes.set(key, mutex);
+  }
+  return mutex;
 }
--- a/packages/core/src/browser/mutex.ts
+++ b/packages/core/src/browser/mutex.ts
@@ -11,6 +11,20 @@
  * window, which forgives the residual cross-tab race the fallback can't cover.
  */
 
+/**
+ * Run `callback` while holding the lock named `key`, resolving to its result.
+ * Only one callback per key runs at a time.
+ */
+export async function runWithMutex<T>(
+  key: string,
+  callback: () => Promise<T>,
+): Promise<T> {
+  const locks = typeof navigator !== "undefined" ? navigator.locks : undefined;
+  return locks !== undefined
+    ? await locks.request(key, callback)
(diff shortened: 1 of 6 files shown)


packages/core/src/browser/mutex.ts
```diff
@@ -0,0 +1,63 @@
...
+      if (next !== undefined) {
+        enqueue(key, next);
+      }
+    });
+  } else {
+    mutex.waiting.push(callback);
+  }
+}
+
+function manualMutex<T>(key: string, callback: () => Promise<T>): Promise<T> {
```
Nicolapps: nit: can you please move `manualMutex` above `getMutex` + `enqueue`? This would make the code slightly easier to read

packages/core/src/browser/mutex.ts
```diff
@@ -0,0 +1,63 @@
+/**
+ * A cross-context mutex used to serialize token refreshes so that concurrent
+ * callers (multiple tabs, or several components racing on mount) collapse to a
+ * single refresh instead of each rotating the refresh token and fighting over
+ * the result.
+ *
+ * When the browser exposes the Web Locks API (`navigator.locks`) it is a true
```
Nicolapps, line 7: Good to know: the web locks API is supported by most browsers (94.21% according to “Can I use…”: https://caniuse.com/mdn-api_lock)

packages/core/src/browser/storage.ts
```diff
@@ -0,0 +1,98 @@
+/**
+ * A framework-agnostic storage abstraction for auth tokens (and any other
```
Nicolapps, line 2: Nice!

packages/core/src/browser/storage.ts
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * The default storage: the browser's `localStorage` when available, otherwise
+ * an {@link InMemoryStorage}. Callers that need cross-tab sessions, React
+ * Native, or per-tab sessions should pass their own storage instead.
+ */
+export function defaultStorage(): TokenStorage {
+  if (typeof window !== "undefined" && window.localStorage) {
+    return window.localStorage;
+  }
+  return new InMemoryStorage();
```
Nicolapps, line 62: Do you think we should `console.warn` here? Or is it too annoying because this code path is likely to be used by user tests?

packages/core/src/browser/sessionManager.ts
```diff
@@ -0,0 +1,308 @@
...
+  authApi: AuthApi;
+  /** Where tokens are persisted. */
+  storage: TokenStorage;
+  /** Namespace for storage keys; typically the deployment URL. */
+  storageNamespace: string;
+  /** Log refresh/lifecycle steps to the console. */
+  verbose?: boolean;
+}
+
+/** The reactive snapshot the React layer subscribes to. */
```
Nicolapps: Nit about the documentation: Is this React-specific? Or should this be something like “user-visible”?

packages/core/src/browser/sessionManager.test.ts
```diff
@@ -0,0 +1,241 @@
...
+  };
+}
+
+/** A resolvable promise, to control refresh timing in the concurrency test. */
+function deferred<T>() {
+  let resolve!: (value: T) => void;
+  const promise = new Promise<T>((r) => {
+    resolve = r;
+  });
+  return { promise, resolve };
```
Nicolapps: FYI you can use `Promise.withResolvers()` to avoid doing this manually

Nicolapps (approved):
Great!

packages/core/src/browser/mutex.ts
dowski: Good suggestion! Also moved `enqueue` above `getMutex` since it also follows more clearly in reading order.

packages/core/src/browser/sessionManager.test.ts
dowski: Oh nice, thanks for the simplification!

packages/core/src/browser/sessionManager.ts
dowski: Thanks, I updated this and also documented the related methods on `AuthClient` which better describe the behavior related to this data.

packages/core/src/browser/storage.ts
dowski: Yeah it'll likely be used by tests.

PR #397, opened 2026-07-10
React client bindings and hooks

dowski:
A React integration built on the framework agnostic parts in #388. Includes everything needed to integrate the core auth functionality with a Convex app using React.

----

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -35,7 +35,8 @@
     "./providers/password/convex.config.js": "./src/components/password/convex.config.ts",
     "./providers/password/*": "./src/components/password/*.ts",
     "./providers/testing/*": "./src/components/testing/*.ts",
-    "./browser": "./src/browser/index.ts"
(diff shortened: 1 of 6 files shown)


packages/core/package.json
```diff
@@ -47,15 +48,27 @@
...
     "vitest": "^4.1.0"
   },
   "peerDependencies": {
-    "convex": "^1.41.0"
+    "convex": "^1.41.0",
+    "react": "^18.0.0 || ^19.0.0"
+  },
+  "peerDependenciesMeta": {
+    "react": {
+      "optional": true
```
Nicolapps, line 71: This looks nice, we might be able to use that for dependencies like `@convex-dev/resend` in the email validation component.

From what I found, it will warn during installation if React is missing (but not fail). And if they have a version of React that doesn’t match our expected range, modern NPM versions will error (unless using `--legacy-peer-deps`).

Nicolapps (approved):

PR #398, opened 2026-07-10
react-minimal example: wire up the client

dowski:
Adds a runnable Vite/React app that uses the `@convex-dev/auth/react` client.

Builds on the React client code in #397 and the anonymous client in #404.

----

### DIFF
--- a/examples/react-minimal/src/main.tsx
+++ b/examples/react-minimal/src/main.tsx
@@ -1,0 +1,22 @@
+import { ConvexAuthProvider } from "@convex-dev/auth/react";
+import { ConvexReactClient } from "convex/react";
+import { StrictMode } from "react";
+import { createRoot } from "react-dom/client";
+import { api } from "../convex/_generated/api";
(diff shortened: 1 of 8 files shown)


examples/react-minimal/src/main.tsx
```diff
@@ -0,0 +1,22 @@
+import { ConvexAuthProvider } from "@convex-dev/auth/react";
+import { ConvexReactClient } from "convex/react";
+import { StrictMode } from "react";
+import { createRoot } from "react-dom/client";
+import { api } from "../convex/_generated/api";
+import { App } from "./App";
+
+const convex = new ConvexReactClient(import.meta.env.VITE_CONVEX_URL as string);
```
Nicolapps: nit: The point of `as string` here is to avoid `undefined`, right? If so, we can simplify it as `import.meta.env.VITE_CONVEX_URL!`.

Nicolapps (approved):

PR #404, opened 2026-07-13
Anonymous provider: React client

dowski:
Add a small React client for the anonymous provider, exported at `@convex-dev/auth/providers/anonymous/react`: `useAnonymousAuth` takes the app's `signInAnonymous` mutation reference, calls it, and hands the resulting `TokenBundle` to the core client's `setSession`.

It will get used in the example app in #398 as well as serve as a simple example itself of what a client API needs to do.

----

### DIFF
--- a/packages/core/src/components/anonymous/react.tsx
+++ b/packages/core/src/components/anonymous/react.tsx
@@ -1,0 +1,60 @@
+/**
+ * React client for the anonymous provider, exported at
+ * `@convex-dev/auth/providers/anonymous/react`.
+ *
+ * A provider's job on the client is to run its own sign-in flow and hand the
+ * resulting {@link TokenBundle} to the core client's `setSession` (see
+ * {@link useAuthActions}). This anonymous provider is the simplest case. It
+ * has a single mutation that directly returns the `TokenBundle`.
+ *
+ * It's useful as an example of the bare minimum required of a provider client.
+ *
+ * @module
+ */
+"use client";
+
+import { useMutation } from "convex/react";
+import { FunctionReference } from "convex/server";
+import { useCallback } from "react";
+import type { TokenBundle } from "../../lib/types";
+import { useAuthActions } from "../../react";
+
+/**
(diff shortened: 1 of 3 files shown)


packages/core/src/components/anonymous/react.tsx
```diff
@@ -0,0 +1,60 @@
...
+ * client authenticates. Passing a `TokenBundle` to `setSession` is the general
+ * pattern that every auth provider should follow.
+ *
+ * ```tsx
+ * import { useAnonymousAuth } from "@convex-dev/auth/providers/anonymous/react";
+ * import { api } from "../convex/_generated/api";
+ *
+ * function SignIn() {
+ *   const { signInAnonymous } = useAnonymousAuth(api.auth.signInAnonymous);
+ *   return <button onClick={() => signInAnonymous()}>Sign in anonymously</button>;
```
Nicolapps, line 47: Presumably, a real usage of this API will want to redirect the user to the next page after they log in, right?

I would expect the example to look something like:

```
return (
  <button
    onClick={async () => {
      await signInAnonymously();
      // redirect the user to the home page
    }}
  >
    Sign in anonymously
  </button>
);
```

Nicolapps: (Also, should the user do error handling here, in case the API call fails?)

Nicolapps (approved):

packages/core/src/components/anonymous/react.tsx
dowski: The `setSession` call is wired up so that `ConvexProviderWithAuth` will pass the token through to the `setAuth` function on the client and the state will change to authenticated. So the page will re-render and the `<Authenticated>` React helper should be enabled (or one can depend on `useConvexAuthActions()`.

But now that I type this out here, it makes sense to include some of that in the documentation! 😆

Nicolapps: I agree that it also works that way.

I think that in practice, most apps would have a redirection step here (because they would have their log in page at another URL than their main page), so it might still be useful

PR #405, opened 2026-07-13
Add framework-agnostic SSR server layer (@convex-dev/auth/server)

dowski:
This is basically the server-side counterpart to `AuthClient` added in #388. These pieces will be used to build SSR integrations, starting with Next.js support in a follow on PR.

----

### DIFF
--- a/packages/core/src/server/cookies.ts
+++ b/packages/core/src/server/cookies.ts
@@ -1,0 +1,71 @@
+/**
+ * A framework-agnostic cookie abstraction for server-side (SSR) auth.
+ *
+ * SSR frameworks each expose cookies differently — Next.js has `NextRequest`/
+ * `NextResponse` cookies in middleware and `next/headers` `cookies()` in route
+ * handlers and server components; other frameworks have their own. This module
+ * defines the small interface the shared {@link ServerAuthSession} needs, so a
+ * framework binding only has to adapt its cookies to {@link CookieStore}.
+ *
+ * Nothing here depends on React, Next.js, or Convex.
+ *
+ * @module
+ */
+
+import { JWT_STORAGE_KEY, REFRESH_TOKEN_STORAGE_KEY } from "../browser/storage";
+
+/** The cookie holding the current access token (a JWT). */
+export const AUTH_JWT_COOKIE = JWT_STORAGE_KEY;
+/** The cookie holding the current (rotating) refresh token. */
+export const AUTH_REFRESH_COOKIE = REFRESH_TOKEN_STORAGE_KEY;
+
+/** Attributes applied when writing a cookie. */
+export interface CookieOptions {
+  /** Whether to hide the cookie from client JS. */
+  httpOnly?: boolean;
+  /** Whether to send only over HTTPS. */
+  secure?: boolean;
+  /** CSRF posture. Defaults to `"lax"`. */
+  sameSite?: "lax" | "strict" | "none";
+  /** Cookie path. Defaults to `"/"`. */
+  path?: string;
+  /** Max age in seconds. */
+  maxAge?: number;
+  /** Absolute expiry. */
+  expires?: Date;
+  /** Cookie domain. */
+  domain?: string;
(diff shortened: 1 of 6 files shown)


packages/core/src/server/jwt.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
+}
+
+/**
+ * Whether a token is expired or within `skewSeconds` of expiry (so the caller
+ * should refresh). A token whose expiry can't be read is treated as expiring.
+ */
+export function isTokenExpiring(
+  token: string,
+  skewSeconds = 10,
```
Nicolapps, line 42: Tim would never 🙃

packages/core/src/server/cookies.ts
```diff
@@ -0,0 +1,71 @@
...
+ * The default attributes for auth cookies: httpOnly (so the tokens never reach
+ * client JS), `secure` in production, `sameSite: "lax"`, rooted at `/`.
+ * Per-cookie lifetime (`expires`) is derived from the token bundle by the
+ * caller.
+ */
+export function defaultCookieOptions(): CookieOptions {
+  return {
+    httpOnly: true,
+    secure: process.env.NODE_ENV === "production",
+    sameSite: "lax",
```
Nicolapps, line 69: Using `"lax"` looks good:

> The simplest method to prevent CSRF is to set the SameSite cookie attribute to Strict or Lax. Strict ensures the cookie is not included in any cross-site requests, while Lax ensures that cookie is only included in GET and GET-like request (HEAD, OPTIONS, TRACE). Using Strict will block the cookie from being sent when the user is redirected from a third-party site, which is usually not the desired behavior. As such, Lax should be the default value.

https://auth.pilcrowonpaper.com/csrf

packages/core/src/server/cookies.ts
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * The default attributes for auth cookies: httpOnly (so the tokens never reach
+ * client JS), `secure` in production, `sameSite: "lax"`, rooted at `/`.
+ * Per-cookie lifetime (`expires`) is derived from the token bundle by the
+ * caller.
+ */
+export function defaultCookieOptions(): CookieOptions {
+  return {
+    httpOnly: true,
+    secure: process.env.NODE_ENV === "production",
```
Nicolapps: This seems to be a little bit dangerous because it makes assumptions about the environment.

Why can’t we default to secure cookies? Is it because they might not work with localhost?

I wonder if we might want to ask the caller to set a value for the `secure` flag, so that each framework can use its standard way of determining if it’s running in dev mode

packages/core/src/server/session.test.ts
```diff
@@ -0,0 +1,195 @@
...
+) {
+  const session = new ServerAuthSession({
+    authApi: {
+      refreshSession: async () => null,
+      signOut: async () => { },
+      ...authApi,
+    },
+    cookies,
+    // Pin "now" so token-expiry math is deterministic.
+    refreshSkewSeconds: 10,
```
Nicolapps: Is this meant to do something else? `refreshSkewSeconds` already defaults to 10, and it doesn’t seem to be pinning `now`

Nicolapps (approved):

packages/core/src/server/cookies.ts
dowski: Good point. You're right about why we can't just default to `true`. Safari doesn't treat `localhost` as secure so it doesn't send cookies with that constraint.

But we can require plumbing the value through from the higher level as maybe that env var won't be set.

packages/core/src/server/session.test.ts
dowski: Thanks. Moved the comment and dropped this no-op override of the default.

packages/core/src/server/cookies.ts
dowski: Thanks, added a reference to that link in the TSDoc.

PR #411, opened 2026-07-16
Add OAuth provider component

erquhart:
Add an OAuth provider component, mounted once per IdP with its own callback route and client credential env bindings.

- Full authorization-code flow: `signIn` mints state and returns the provider URL, the component's HTTP callback exchanges the code and mints a one-time ticket, `redeem` completes sign-in through the core (accounts, sessions, tokens)
- Supports OIDC (Google) and plain OAuth (GitHub) via `userInfoEndpoints` + a `profile` mapping
- Hardened: encrypted ticket payloads, state binding at redemption (login CSRF), strict id_token validation, PKCE, deploy-time config validation, fetch timeouts, callback failures always redirect back to the app
- Example app with a full React client: token lifecycle (localStorage, refresh, sign-out), Google + GitHub buttons, tests

Not yet:

- Cleanup of expired rows (authorization requests, tickets)
- Rate limiting on the public sign-in surface
- `form_post` callbacks (Apple)
- Callback HTTP test coverage
- Hardening

Note that the inline client in the example app works but didn't get a lot of scrutiny, added for completeness and testability.

----

### DIFF
--- a/examples/react-oauth/README.md
+++ b/examples/react-oauth/README.md
@@ -1,0 +1,19 @@
+# react-oauth example
+
+An in-repo Convex backend that mounts the `@convex-dev/auth` core component
+together with its oauth provider. It exists as an example of wiring oauth
+into an application and for integration testing the oauth provider end to end.
+
+## Generate code / run
+
+Run the example from its own directory.
+
+```bash
+cd examples/react-oauth
+npx convex dev --once    # provisions a deployment, generates convex/_generated
+npx generate-auth-keys   # sets AUTH_PRIVATE_KEY + AUTH_JWKS on the deployment
+```
+
+## Test usage
+
+The tests defined in the example are run along with `pnpm test` in the repo root.
--- a/examples/react-oauth/convex/auth.ts
+++ b/examples/react-oauth/convex/auth.ts
@@ -16,9 +16,9 @@ export const {
   component: components.core,
   providers: [
     provider(Oauth, {
-      component: components.authOauth,
       providers: {
         google: {
+          component: components.oauthGoogle,
           clientId: process.env.AUTH_GOOGLE_CLIENT_ID ?? "",
           authorizationEndpoint: "https://accounts.google.com/o/oauth2/v2/auth",
           scopes: ["openid", "email", "profile"],
--- a/examples/react-oauth/convex/auth.ts
+++ b/examples/react-oauth/convex/auth.ts
@@ -3,28 +3,25 @@ import { provider, setupCore } from "@convex-dev/auth/core/setup";
 import { Oauth } from "@convex-dev/auth/providers/oauth/setup";
 
 // The core owns sessions, accounts, and JWT minting. It calls back into our
-// `createOrUpdateUser` on every sign-in from a provider. The oauth provider
-// exposes `signInOauth`, which records an authorization request and returns
-// the provider authorization URL for the client to navigate to.
+// `createOrUpdateUser` on every sign-in from a provider. Each Oauth instance
+// covers one IdP and exposes `signIn`, which records an authorization
+// request and returns the provider authorization URL for the client to
+// navigate to.
 export const {
   signOut,
   refreshSession,
   providers: {
-    oauth: { signInOauth },
+    google: { signIn: signInGoogle },
   },
 } = setupCore({
   component: components.core,
   providers: [
-    provider(Oauth, {
-      providers: {
-        google: {
-          component: components.oauthGoogle,
-          authorizationEndpoint: "https://accounts.google.com/o/oauth2/v2/auth",
-          tokenEndpoint: "https://oauth2.googleapis.com/token",
-          scopes: ["openid", "email", "profile"],
-          pkce: true,
-        },
-      },
+    provider(Oauth("google"), {
+      component: components.oauthGoogle,
+      authorizationEndpoint: "https://accounts.google.com/o/oauth2/v2/auth",
+      tokenEndpoint: "https://oauth2.googleapis.com/token",
+      scopes: ["openid", "email", "profile"],
+      pkce: true,
       allowedRedirectOrigins: ["http://localhost:5173"],
     }),
   ],
--- a/examples/react-oauth/convex/auth.ts
+++ b/examples/react-oauth/convex/auth.ts
@@ -3,19 +3,29 @@ import { provider, setupCore } from "@convex-dev/auth/core/setup";
 import { Oauth } from "@convex-dev/auth/providers/oauth/setup";
 
 // The core owns sessions, accounts, and JWT minting. It calls back into our
-// `createOrUpdateUser` on every sign-in from a provider. The password provider
-// exposes the sign up / sign in actions, keyed to an opaque app user id.
+// `createOrUpdateUser` on every sign-in from a provider. The oauth provider
+// exposes `signInOauth`, which records an authorization request and returns
+// the provider authorization URL for the client to navigate to.
 export const {
   signOut,
   refreshSession,
   providers: {
-    oauth: { signIn },
+    oauth: { signInOauth },
   },
 } = setupCore({
   component: components.core,
   providers: [
     provider(Oauth, {
       component: components.authOauth,
+      providers: {
+        google: {
+          clientId: process.env.AUTH_GOOGLE_CLIENT_ID ?? "",
+          authorizationEndpoint: "https://accounts.google.com/o/oauth2/v2/auth",
+          scopes: ["openid", "email", "profile"],
+          pkce: true,
+        },
+      },
+      allowedRedirectOrigins: ["http://localhost:5173"],
     }),
   ],
 }).attachUserCallback(internal.users.createOrUpdateUser);
--- a/examples/react-oauth/convex/auth.ts
+++ b/examples/react-oauth/convex/auth.ts
@@ -20,6 +20,7 @@ export const {
       component: components.oauthGoogle,
       authorizationEndpoint: "https://accounts.google.com/o/oauth2/v2/auth",
       tokenEndpoint: "https://oauth2.googleapis.com/token",
+      issuer: "https://accounts.google.com",
       scopes: ["openid", "email", "profile"],
       pkce: true,
       allowedRedirectOrigins: ["http://localhost:5173"],
--- a/examples/react-oauth/convex/auth.ts
+++ b/examples/react-oauth/convex/auth.ts
@@ -1,0 +1,21 @@
+import { components, internal } from "./_generated/api";
+import { provider, setupCore } from "@convex-dev/auth/core/setup";
+import { Oauth } from "@convex-dev/auth/providers/oauth/setup";
+
+// The core owns sessions, accounts, and JWT minting. It calls back into our
+// `createOrUpdateUser` on every sign-in from a provider. The password provider
+// exposes the sign up / sign in actions, keyed to an opaque app user id.
+export const {
+  signOut,
+  refreshSession,
+  providers: {
+    oauth: { signIn },
+  },
+} = setupCore({
+  component: components.core,
+  providers: [
+    provider(Oauth, {
+      component: components.authOauth,
+    }),
+  ],
+}).attachUserCallback(internal.users.createOrUpdateUser);
--- a/examples/react-oauth/convex/auth.ts
+++ b/examples/react-oauth/convex/auth.ts
@@ -7,11 +7,15 @@ import { Oauth } from "@convex-dev/auth/providers/oauth/setup";
 // covers one IdP and exposes `signIn`, which records an authorization
 // request and returns the provider authorization URL for the client to
 // navigate to.
+/** A GitHub `/user/emails` entry, the fields the profile mapping reads. */
+type GithubEmail = { email: string; primary: boolean; verified: boolean };
+
 export const {
   signOut,
   refreshSession,
   providers: {
     google: { signIn: signInGoogle, redeem: redeemGoogle },
+    github: { signIn: signInGithub, redeem: redeemGithub },
   },
 } = setupCore({
   component: components.core,
@@ -25,6 +29,34 @@ export const {
       issuer: "https://accounts.google.com",
       scopes: ["openid", "email", "profile"],
       pkce: true,
+      allowedRedirectOrigins: ["http://localhost:5173"],
+    }),
+    // GitHub is plain OAuth (no id_token), so identity comes from userinfo
+    // endpoints and a profile mapping. No `pkce`: GitHub OAuth apps silently
+    // ignore it.
+    provider(Oauth("github"), {
+      component: components.oauthGithub,
+      // Must match the mount in convex.config.ts.
+      httpPrefix: "/oauth/github",
+      authorizationEndpoint: "https://github.com/login/oauth/authorize",
+      tokenEndpoint: "https://github.com/login/oauth/access_token",
+      scopes: ["read:user", "user:email"],
+      userInfoEndpoints: {
+        user: "https://api.github.com/user",
+        emails: "https://api.github.com/user/emails",
+      },
+      profile: (_claims, userInfoResponses) => {
+        const user = userInfoResponses?.user;
+        const emails: GithubEmail[] = userInfoResponses?.emails ?? [];
+        const best =
+          emails.find((entry) => entry.primary && entry.verified) ??
+          emails.find((entry) => entry.verified);
+        return {
+          id: String(user.id),
+          email: best?.email ?? user.email,
+          name: user.name ?? user.login,
+        };
+      },
       allowedRedirectOrigins: ["http://localhost:5173"],
     }),
   ],
--- a/examples/react-oauth/convex/auth.ts
+++ b/examples/react-oauth/convex/auth.ts
@@ -11,7 +11,7 @@ export const {
   signOut,
   refreshSession,
   providers: {
-    google: { signIn: signInGoogle },
+    google: { signIn: signInGoogle, redeem: redeemGoogle },
   },
 } = setupCore({
   component: components.core,
--- a/examples/react-oauth/convex/auth.ts
+++ b/examples/react-oauth/convex/auth.ts
@@ -1,4 +1,5 @@
 import { components, internal } from "./_generated/api";
+import { env } from "./_generated/server";
 import { provider, setupCore } from "@convex-dev/auth/core/setup";
 import { Oauth } from "@convex-dev/auth/providers/oauth/setup";
 
@@ -19,7 +20,7 @@ export const {
       providers: {
         google: {
           component: components.oauthGoogle,
-          clientId: process.env.AUTH_GOOGLE_CLIENT_ID ?? "",
+          clientId: env.AUTH_GOOGLE_CLIENT_ID,
           authorizationEndpoint: "https://accounts.google.com/o/oauth2/v2/auth",
           scopes: ["openid", "email", "profile"],
           pkce: true,
--- a/examples/react-oauth/convex/auth.ts
+++ b/examples/react-oauth/convex/auth.ts
@@ -1,5 +1,4 @@
 import { components, internal } from "./_generated/api";
-import { env } from "./_generated/server";
 import { provider, setupCore } from "@convex-dev/auth/core/setup";
 import { Oauth } from "@convex-dev/auth/providers/oauth/setup";
 
@@ -20,8 +19,8 @@ export const {
       providers: {
         google: {
           component: components.oauthGoogle,
-          clientId: env.AUTH_GOOGLE_CLIENT_ID,
           authorizationEndpoint: "https://accounts.google.com/o/oauth2/v2/auth",
+          tokenEndpoint: "https://oauth2.googleapis.com/token",
           scopes: ["openid", "email", "profile"],
           pkce: true,
         },
--- a/examples/react-oauth/convex/auth.ts
+++ b/examples/react-oauth/convex/auth.ts
@@ -18,6 +18,8 @@ export const {
   providers: [
     provider(Oauth("google"), {
       component: components.oauthGoogle,
+      // Must match the mount in convex.config.ts.
+      httpPrefix: "/oauth/google",
       authorizationEndpoint: "https://accounts.google.com/o/oauth2/v2/auth",
       tokenEndpoint: "https://oauth2.googleapis.com/token",
       issuer: "https://accounts.google.com",
--- a/examples/react-oauth/convex/oauth.test.ts
+++ b/examples/react-oauth/convex/oauth.test.ts
@@ -27,13 +27,14 @@ async function setup() {
   );
   // convex-test doesn't emulate per-mount env binding renames (convex.config
   // maps AUTH_GOOGLE_CLIENT_ID onto the component's CLIENT_ID), so stub the
-  // component-side names directly.
+  // component-side names directly; both mounts see the same credentials.
   vi.stubEnv("CLIENT_ID", "test-client-id");
   vi.stubEnv("CLIENT_SECRET", "test-client-secret");
 
   const t = convexTest(schema, modules);
   registerCore(t);
   registerOauthProvider(t, "oauthGoogle");
+  registerOauthProvider(t, "oauthGithub");
   return t;
 }
 
@@ -79,6 +80,34 @@ describe("oauth", () => {
   test("redeem returns null for an unknown code", async () => {
     const t = await setup();
     const result = await t.mutation(api.auth.redeemGoogle, {
+      code: "not-a-real-code",
+      state: "not-a-real-state",
+    });
+    expect(result).toBeNull();
+  });
+
+  test("github signIn returns its own authorization URL without PKCE", async () => {
+    const t = await setup();
+    const { redirect, state } = await t.mutation(api.auth.signInGithub, {
+      redirectTo: "http://localhost:5173/",
+    });
+    const url = new URL(redirect);
+    expect(`${url.origin}${url.pathname}`).toBe(
+      "https://github.com/login/oauth/authorize",
+    );
+    expect(url.searchParams.get("redirect_uri")).toBe(
+      "https://example.convex.site/oauth/github/callback",
+    );
+    expect(url.searchParams.get("scope")).toBe("read:user user:email");
+    expect(url.searchParams.get("state")).toBe(state);
+    // The github provider is configured without `pkce`.
+    expect(url.searchParams.get("code_challenge")).toBeNull();
+    expect(url.searchParams.get("code_challenge_method")).toBeNull();
+  });
+
+  test("github redeem returns null for an unknown code", async () => {
+    const t = await setup();
+    const result = await t.mutation(api.auth.redeemGithub, {
       code: "not-a-real-code",
       state: "not-a-real-state",
     });
--- a/examples/react-oauth/convex/oauth.test.ts
+++ b/examples/react-oauth/convex/oauth.test.ts
@@ -3,12 +3,10 @@ import { afterEach, describe, expect, test, vi } from "vitest";
 import { exportJWK, exportPKCS8, generateKeyPair } from "jose";
 import { api } from "./_generated/api.js";
 import { registerCore } from "@convex-dev/auth/providers/testing/core";
-import { registerPasswordProvider } from "@convex-dev/auth/providers/testing/password";
+import { registerOauthProvider } from "@convex-dev/auth/providers/testing/oauth";
 import schema from "./schema.js";
 
 const modules = import.meta.glob("./**/*.ts");
-
-const PASSWORD = "correct horse battery staple"; // 28 chars, valid
 
 async function setup() {
   // The core signs JWTs from these env vars (see core/public.ts). Mint a real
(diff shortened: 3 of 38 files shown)


examples/react-oauth/convex/auth.ts
```diff
@@ -0,0 +1,63 @@
...
+  providers: [
+    provider(Oauth("google"), {
+      component: components.oauthGoogle,
+      // Must match the mount in convex.config.ts.
+      httpPrefix: "/oauth/google",
+      authorizationEndpoint: "https://accounts.google.com/o/oauth2/v2/auth",
+      tokenEndpoint: "https://oauth2.googleapis.com/token",
+      issuer: "https://accounts.google.com",
+      scopes: ["openid", "email", "profile"],
+      pkce: true,
```
dowski, line 31: Would someone ever want to set this to false?

examples/react-oauth/convex/auth.ts
```diff
@@ -0,0 +1,63 @@
...
+      component: components.oauthGithub,
+      // Must match the mount in convex.config.ts.
+      httpPrefix: "/oauth/github",
+      authorizationEndpoint: "https://github.com/login/oauth/authorize",
+      tokenEndpoint: "https://github.com/login/oauth/access_token",
+      scopes: ["read:user", "user:email"],
+      userInfoEndpoints: {
+        user: "https://api.github.com/user",
+        emails: "https://api.github.com/user/emails",
+      },
```
dowski, line 47: Many of these config values are specific to GitHub and won't vary, right? I think it might be nice to internalize a lot of this for the few providers we directly support, rather than it being something that has to be expressed at this config layer.

examples/react-oauth/convex/auth.ts
```diff
@@ -0,0 +1,63 @@
...
+        const emails: GithubEmail[] = userInfoResponses?.emails ?? [];
+        const best =
+          emails.find((entry) => entry.primary && entry.verified) ??
+          emails.find((entry) => entry.verified);
+        return {
+          id: String(user.id),
+          email: best?.email ?? user.email,
+          name: user.name ?? user.login,
+        };
+      },
```
dowski, line 59: This also feels like it could be internal to the GitHub provider.

examples/react-oauth/convex/oauth.test.ts
```diff
@@ -0,0 +1,116 @@
...
+}
+
+afterEach(() => {
+  vi.unstubAllEnvs();
+});
+
+// The full callback flow (code exchange, ticket mint) runs over the
+// component's HTTP route, which app-level convex-test doesn't serve; it's
+// covered by the component's own tests. These tests cover the app-facing
+// mutations.
```
dowski, line 48: Ah interesting. I wonder if we should add that caveat to https://docs.convex.dev/testing/convex-test#http-actions, or actually make it work.

examples/react-oauth/convex/users.ts
```diff
@@ -0,0 +1,59 @@
...
+ * The signed-in user's document, or `null` when unauthenticated. The JWT's
+ * subject is the app user id minted by `createOrUpdateUser`.
+ */
+export const me = query({
+  args: {},
+  returns: v.union(
+    v.null(),
+    v.object({
+      _id: v.id("users"),
+      _creationTime: v.number(),
```
dowski, line 14: Are these useful to return? Maybe just strip it down to just returning an object with `email`?

examples/react-oauth/src/App.tsx
```diff
@@ -0,0 +1,140 @@
...
+const GitHubIcon = (): ReactNode => (
+  <svg
+    className="h-5 w-5 shrink-0"
+    viewBox="0 0 24 24"
+    fill="currentColor"
+    aria-hidden="true"
+  >
+    <path d="M12 .5C5.37.5 0 5.87 0 12.5c0 5.3 3.44 9.8 8.21 11.39.6.11.82-.26.82-.58 0-.29-.01-1.04-.02-2.05-3.34.73-4.04-1.61-4.04-1.61-.55-1.39-1.34-1.76-1.34-1.76-1.09-.75.08-.73.08-.73 1.21.09 1.85 1.24 1.85 1.24 1.07 1.84 2.81 1.31 3.5 1 .11-.78.42-1.31.76-1.61-2.67-.3-5.47-1.34-5.47-5.95 0-1.31.47-2.39 1.24-3.23-.13-.3-.54-1.53.12-3.18 0 0 1.01-.32 3.3 1.23a11.5 11.5 0 0 1 6 0c2.29-1.55 3.3-1.23 3.3-1.23.66 1.65.25 2.88.12 3.18.77.84 1.24 1.92 1.24 3.23 0 4.62-2.81 5.64-5.49 5.94.43.37.81 1.1.81 2.22 0 1.6-.01 2.9-.01 3.29 0 .32.22.7.83.58A12.01 12.01 0 0 0 24 12.5C24 5.87 18.63.5 12 .5z" />
+  </svg>
+);
```
dowski, line 38: Consider putting the SVGs in their own files to simplify the example code.

examples/react-oauth/src/App.tsx
```diff
@@ -0,0 +1,140 @@
...
+        <p className="mb-4 rounded-lg bg-red-50 px-4 py-2.5 text-sm text-red-700">
+          {flowError}
+        </p>
+      )}
+      <div className="flex flex-col gap-3">
+        {PROVIDERS.map((provider) => (
+          <button
+            key={provider.id}
+            type="button"
+            onClick={() => void signIn(provider.id)}
```
dowski, line 90: I think it would be nice to avoid a generic `signIn` function. Flows are different enough across providers that one function has to do a lot to make everything work.

The Anonymous provider offers its own API (as an example).

examples/react-oauth/src/auth.tsx
dowski, line 1: Is this a file that we'd expect app authors to create on their own? Or should this functionality be provided by Convex Auth in a client library?

examples/react-oauth/src/auth.tsx
```diff
@@ -0,0 +1,324 @@
...
+  );
+  const actions = useMemo(
+    () => ({ signIn, signOut, flowError }),
+    [signIn, signOut, flowError],
+  );
+
+  return (
+    <AuthStateContext.Provider value={authState}>
+      <AuthActionsContext.Provider value={actions}>
+        <ConvexProviderWithAuth client={client} useAuth={useAuthState}>
```
dowski, line 318: I think most of this should wind up being handled/wired up in `packages/core/src/react/client.tsx`.

I guess maybe the point of this is to just show conceptually how it could be wired up?

packages/core/src/components/oauth/crypto.ts
```diff
@@ -0,0 +1,119 @@
...
+  const binary = atob(base64);
+  return Uint8Array.from(binary, (c) => c.charCodeAt(0));
+}
+
+/**
+ * Decode a JWT's payload without verifying its signature. Only safe for
+ * tokens received directly from the issuer over TLS (e.g. an id_token from
+ * the token exchange), where transport authenticates the issuer.
+ */
+export function decodeJwtPayload(jwt: string): Record<string, unknown> {
```
dowski, line 54: Maybe name this something like `unverifiedJwtPayload`. That way it's more clear what the function is doing.

packages/core/src/components/oauth/crypto.ts
```diff
@@ -0,0 +1,119 @@
...
+    "encrypt",
+    "decrypt",
+  ]);
+}
+
+/**
+ * Encrypt a ticket payload with a key derived from the raw one-time token.
+ * The token is never persisted (only its hash is), so database access alone
+ * cannot decrypt the result; the key preimage travels only in the callback
+ * redirect URL. Returns base64url(iv || ciphertext).
```
dowski, line 82: It's a bit odd for the doc to refer to internal variable names.

packages/core/src/components/oauth/crypto.ts
```diff
@@ -0,0 +1,119 @@
...
+  combined.set(ciphertext, iv.length);
+  return base64UrlEncode(combined);
+}
+
+/**
+ * Decrypt {@link encryptWithToken} output. Throws when the token is wrong
+ * or the payload was tampered with (AES-GCM authenticates).
+ */
+export async function decryptWithToken(
+  token: string,
```
dowski, line 108: Maybe name this base64EncodedToken or encodedToken at least?

packages/core/src/components/oauth/crypto.ts
```diff
@@ -0,0 +1,119 @@
...
+  return await crypto.subtle.importKey("raw", keyBytes, "AES-GCM", false, [
+    "encrypt",
+    "decrypt",
+  ]);
+}
+
+/**
+ * Encrypt a ticket payload with a key derived from the raw one-time token.
+ * The token is never persisted (only its hash is), so database access alone
+ * cannot decrypt the result; the key preimage travels only in the callback
```
dowski, line 81: I'm not familiar with crypto jargon I guess. "preimage" doesn't mean anything to me, but maybe that's okay. 😆

packages/core/src/components/oauth/crypto.ts
```diff
@@ -0,0 +1,119 @@
...
+function base64UrlEncode(bytes: Uint8Array): string {
+  const binary = Array.from(bytes, (b) => String.fromCharCode(b)).join("");
+  return btoa(binary)
+    .replace(/\+/g, "-")
+    .replace(/\//g, "_")
+    .replace(/=+$/, "");
+}
+
+/** SHA-256 hex digest, used so raw state and one-time tokens are never persisted. */
+export async function sha256Hex(value: string): Promise<string> {
```
dowski, line 16: We have a version of this function in `packages/core/src/components/core/crypto.ts`. Do you think it's worth putting in one place that both core and OAuth use?

packages/core/src/components/oauth/http.ts
```diff
@@ -0,0 +1,307 @@
...
+    status: 302,
+    headers: { Location: location },
+  });
+}
+
+/**
+ * 302 back to the app's `redirectTo` with outcome params. Stale
+ * `code`/`error` params from a previous attempt are removed first, so a
+ * `redirectTo` derived from the app's current URL can't carry a
+ * contradictory outcome.
```
dowski, line 33: How would stale params show up?

packages/core/src/components/oauth/http.ts
```diff
@@ -0,0 +1,307 @@
...
+  callbackUrl: string;
+  codeVerifier: string | undefined;
+}): Promise<{ idToken: string | undefined; accessToken: string | undefined }> {
+  const body = new URLSearchParams({
+    grant_type: "authorization_code",
+    code: args.code,
+    client_id: env.CLIENT_ID,
+    client_secret: env.CLIENT_SECRET,
+    // Must byte-match the redirect_uri from the authorization request, so
+    // it comes off the request row rather than being rebuilt here.
```
dowski, line 99: What's the request row? Are the `args` sourced from the database table?

packages/core/src/components/oauth/http.ts
```diff
@@ -0,0 +1,307 @@
...
+ * Decode an id_token and check the claims that guard against config mixups
+ * and issuer collisions: the token was minted by the configured issuer
+ * (`iss`; required whenever an id_token comes back, because `sub` is only
+ * unique within an issuer and a shared or multi-tenant token endpoint can
+ * serve several), for this OAuth app alone (`aud` must be exactly
+ * CLIENT_ID; multi-audience tokens are rejected because no other audience
+ * is trusted, and `azp` must match when present), and isn't expired
+ * (`exp`). Signature verification is deliberately skipped: the token came
+ * directly from the provider's token endpoint over TLS, which OIDC Core
+ * sanctions in place of checking the signature.
```
dowski, line 137: This is quite a wall of text and jargon (thanks Claude). Can you reformat it as a list maybe that more clearly lays out the checks that are performed?

packages/core/src/components/oauth/http.ts
```diff
@@ -0,0 +1,307 @@
...
+ */
+function validateIdToken(
+  idToken: string,
+  expectedIssuer: string | undefined,
+): Record<string, unknown> {
+  if (expectedIssuer === undefined) {
+    throw new Error(
+      "The provider returned an id_token but no issuer is configured; set `issuer` in the provider options so the token can be validated",
+    );
+  }
```
dowski, line 147: Is there any way this can be prevented structurally/via strong typing?

packages/core/src/components/oauth/http.ts
```diff
@@ -0,0 +1,307 @@
...
+  return claims;
+}
+
+/**
+ * Fetch each configured userinfo endpoint with the access token. All
+ * popular providers accept the same shape — GET with a bearer token — with
+ * provider variation living in the endpoint URL's query params.
+ */
+async function fetchUserInfo(
```
dowski, line 170: This isn't required by all providers though, is it? I thought Google included this info w/out an extra fetch.

packages/core/src/components/oauth/provider.ts
```diff
@@ -0,0 +1,181 @@
...
+  ),
+  handler: async (ctx, args) => {
+    const request = await ctx.db
+      .query("authorizationRequests")
+      .withIndex("stateHash", (q) => q.eq("stateHash", args.stateHash))
+      .unique();
+    if (request === null) {
+      return null;
+    }
+    await ctx.db.delete("authorizationRequests", request._id);
```
dowski, line 94: I wonder if this should handle multiple results from the `stateHash`. It would indicate a client bug (multiple requests to `createAuthorizationRequest` with the same input), and would be good to clean up if encountered.

That or prevent multiple `authorizationRequests` from being created for the same `stateHash`.

packages/core/src/components/oauth/schema.ts
```diff
@@ -0,0 +1,86 @@
...
+    /** Hash of the server-minted state. The raw value is never stored. */
+    stateHash: v.string(),
+    /** Post-login destination, validated against allowed redirects at sign-in. */
+    redirectTo: v.string(),
+    /**
+     * The OAuth `redirect_uri`, built app-side at sign-in from
+     * `CONVEX_SITE_URL` plus the mount's `httpPrefix` (the component can't
+     * see system env vars). Stored so the code exchange presents the
+     * byte-identical value, as OAuth requires.
+     */
```
dowski, line 23: I saw something like this in a comment elsewhere. `CONVEX_SITE_URL` in a component is the parent `CONVEX_SITE_URL` plus the mount point that the parent configured it with. It's possible that could simplify things.

https://docs.convex.dev/components/authoring#http-actions

packages/core/src/components/oauth/setup.ts
```diff
@@ -0,0 +1,413 @@
...
+  /**
+   * The `httpPrefix` this IdP's component is mounted under in
+   * convex.config.ts, e.g. `"/oauth/google"`. Must match that mount: the
+   * OAuth `redirect_uri` is built app-side as
+   * `${CONVEX_SITE_URL}${httpPrefix}/callback` because the component itself
+   * can't see the system env var (a typed-env component's `process.env`
+   * only contains its bound vars). A mismatch surfaces on the first sign-in
+   * attempt as the provider rejecting an unregistered redirect URI.
+   */
+  httpPrefix: string;
```
dowski, line 49: Yeah, see my other comment re: this and possible simplification.

packages/core/src/components/oauth/setup.ts
```diff
@@ -0,0 +1,413 @@
...
+ * `<site-url><httpPrefix>/callback` as the redirect URI with each provider.
+ *
+ * The callback only accepts GET redirects. Providers that POST it
+ * (`response_mode=form_post`, notably Apple when name/email scopes are
+ * requested) are not supported yet.
+ */
+export function Oauth<const N extends string>(name: N) {
+  return defineProvider({
+    name,
+    setup: (helpers, options: OauthOptions) => {
```
dowski, line 192: It might be better to name the returned functions `startSignIn` and `completeSignIn` to indicate their direct purpose.

dowski (changes requested):
Nice progress!

I think we'll want to do further review/submission in smaller chunks. WDYT?

examples/react-oauth/convex/auth.ts
erquhart: Yes, its per provider and some don't support it. Most users won't be filling out this object, they'll use a google() helper or something that we provide, and that helper is what sets properties like this based on what the provider supports.

examples/react-oauth/convex/auth.ts
erquhart: The idea is to have a generic api so folks could wire up any oauth provider (this interface supports the bulk of them), but we have wrappers for the ones we officially support - they wouldn't enter all this info, it'd just be github() and maybe they pass in scopes once we do more than strictly authentication.

examples/react-oauth/convex/auth.ts
erquhart: yeah, would be wrapped in a github helper. the importance of this surface existing at the app level is a dev trying to do something we haven't thought of can just do it. Super hackable.

examples/react-oauth/convex/oauth.test.ts
erquhart: I'd lean make it work but can add a note for now

examples/react-oauth/convex/users.ts
erquhart: Agree, this is just a dummy client for review purposes, needs another round to be a proper public example.

examples/react-oauth/src/auth.tsx
erquhart: I should have been more prominent with my note in the OP - this client was thoughtlessly thrown together review of the component itself. The example app is not an actual example of the client api surfaces we'll actually want. Just shows the raw inputs/outputs at the client interaction layer. I would probably do everything differently - an upcoming PR can handle that.

examples/react-oauth/src/auth.tsx
erquhart: Yep, exactly - this is purely to show that the backend is functional, and the example runs so if someone wanted to see the backend work in action for review purposes they could.

packages/core/src/components/oauth/crypto.ts
erquhart: If the core one is the argon2id approach, it may be heavier than we need for OTT/state. The argon2id hash can take like a second if I remember correctly, which is worth it for one time password creation or update, but on every login a second adds up. Need to verify the time difference first to make sure this is accurate.

packages/core/src/components/oauth/crypto.ts
erquhart: Good idea, will do

packages/core/src/components/oauth/crypto.ts
erquhart: lol I'm not familiar with the term either but it made sense in context, its just the encryption key itself. I guess I don't know what "preimage" is meant to add here honestly. will update

packages/core/src/components/oauth/crypto.ts
erquhart: good catch

packages/core/src/components/oauth/crypto.ts
erquhart: the encrypted part is what gets encoded, we'd need to rename the encrypt function too since it encodes, I was kind of thinking of them as a pair so the encoding is just something the pair does. Maybe encrypt/decryptAndEncodeWithToken would be better?

packages/core/src/components/oauth/http.ts
erquhart: Yes there's an authorizationRequest row that's created when the attempt begins and captures everything thats needed for the exchange.

packages/core/src/components/oauth/http.ts
erquhart: Would only happen if the client failed to strip an error from the url, then the user tries logging in again, so the risk is they come back with both an ott and error on the second attempt. Our client would strip so this is just backup (eg., someone could build their own client)

packages/core/src/components/oauth/crypto.ts
erquhart: Ah I was wrong, the core one is just doing the same thing as this. If we share it what do you think of pulling it out to a core/src/lib directory?

packages/core/src/components/oauth/http.ts
erquhart: agree, will do

packages/core/src/components/oauth/http.ts
erquhart: We can improve things with that and get a type error for proper config, but we still need this check in case the runtime reality is that we get an id token and no issuer was configured.

packages/core/src/components/oauth/http.ts
erquhart: It's not required by all providers - it only runs if userInfoEndpoints is populated in the oauth provider config.

erquhart: If a provider is oidc (like Google) then we get claims and no userInfoEndpoints are necessary. With GitHub (non-oidc), there are no claims so the endpoints are where we fetch from and the profile mapping function in the provider config processes the returned values. Not GitHub specific, and also useful for some that do have oidc but where additional data fetches are common at signin.

packages/core/src/components/oauth/schema.ts
erquhart: Oooh only within the http action! Saw that it wasn't defined in the component and did this instead - will see if we can thread that through, would be much better.

packages/core/src/components/oauth/provider.ts
erquhart: It's not an actual risk, if the client attempts signIn multiple times each sign in runs through a component function that generates new hash - would require a hash collision. Will def add if preferred though

packages/core/src/components/oauth/setup.ts
erquhart: Yeah that's much better, will do

examples/react-oauth/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@convex-dev/auth-example-react-oauth",
```
Nicolapps, line 2: Can you please use the same simplified structure as for the other examples?

This includes:
- package name = `example-react-oauth`
- styling: minimal `index.css`, no tailwind
- simplified HTML structure, so that we can embed it in docs and make the components easy to copy + paste

examples/react-oauth/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "version": "0.0.0",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "dev": "concurrently -n vite,convex -c green,blue \"vite\" \"convex dev --tail-logs\"",
```
Nicolapps, line 7: We can use `--start` so that we can skip the concurrently dep:

```suggestion
    "dev": "convex dev --start vite",
```

Also, I think that `--tail-logs` without any extra parameter doesn’t change anything compared to now providing it.

examples/react-oauth/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "build": "vite build",
+    "preview": "vite preview",
```
Nicolapps, line 9: To match the other templates and make sure the commands handle Convex code correctly:

```suggestion
    "build": "vite build && tsc -p convex --noEmit",
    "typecheck": "tsc --noEmit && tsc -p convex --noEmit",
```

examples/react-oauth/convex/oauth.test.ts
Nicolapps: IIUC the problem is that convex-test doesn’t support the new `httpPrefix` parameter for components, is that right?

examples/react-oauth/convex/users.ts
```diff
@@ -0,0 +1,59 @@
...
+  },
+  returns: v.id("users"),
+  handler: async (ctx, args) => {
+    if (args.userId !== null) {
+      const existing = ctx.db.normalizeId("users", args.userId);
+      if (existing === null) {
+        throw new Error(`Unknown user id: ${args.userId}`);
+      }
+      return existing;
+    }
```
Nicolapps, line 52: I never realized that we are doing this check. I wonder if we could simplify it in all demos to

```suggestion
    if (args.userId !== null) {
      return args.userId;
    }
```

This might require us to fix the type of `args.userId` (it should already be typed as `v.id("users")`)

examples/react-oauth/README.md
```diff
@@ -0,0 +1,19 @@
...
+into an application and for integration testing the oauth provider end to end.
+
+## Generate code / run
+
+Run the example from its own directory.
+
+```bash
+cd examples/react-oauth
+npx convex dev --once    # provisions a deployment, generates convex/_generated
+npx generate-auth-keys   # sets AUTH_PRIVATE_KEY + AUTH_JWKS on the deployment
```
Nicolapps, line 14:
```suggestion
npx @convex-dev/auth
```

I removed the standalone `npx generate-auth-keys` export recently, but maybe we want to restore it?

examples/react-oauth/src/App.tsx
Nicolapps: I fully agree. `signIn` makes it also more difficult to implement more complex login flows.

packages/core/src/components/oauth/constants.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Path the component's HTTP router serves the provider callback under,
+ * relative to the mount's `httpPrefix`. Also the suffix of the redirect URI
+ * built app-side and registered with the provider, so every use must
+ * byte-match.
+ */
+export const CALLBACK_PATH = "/callback";
```
Nicolapps, line 7: It sounds like the component only needs to expose a single HTTP route. If that is the case, could we simplify the path so that we can remove `/callback`?

examples/react-oauth/src/auth.tsx
```diff
@@ -0,0 +1,324 @@
...
+    if (bundle.refreshTokenExpiresAt <= Date.now()) {
+      localStorage.removeItem(TOKENS_KEY);
+      return null;
+    }
+    return bundle;
+  } catch {
+    localStorage.removeItem(TOKENS_KEY);
+    return null;
+  }
+}
```
Nicolapps, line 56: I think it might be useful if our client library had a general utility that could be used to store any kind of non-auth sessions (e.g. for OAuth, password reset flows, multi-step login flows…). We don't need to change anything for now, but it can be interesting to consider

examples/react-oauth/src/auth.tsx
```diff
@@ -0,0 +1,324 @@
...
+
+/** User-facing messages for the callback's normalized `error` param. */
+const FLOW_ERROR_MESSAGES: Record<string, string> = {
+  access_denied: "Sign-in was cancelled.",
+  expired: "Sign-in took too long. Please try again.",
+  oauth_error: "Something went wrong during sign-in. Please try again.",
+};
```
Nicolapps, line 139: For the real client implementation, could you please consider providing a way for users to:
- customize error messages
- make sure they handle error messages exhaustively

Maybe we could ensure at the type level that users are handling every possible error cases. We could try doing something similar to the password component.

examples/react-oauth/src/auth.tsx
```diff
@@ -0,0 +1,324 @@
...
+ * Flow: `signIn` asks the server for an authorization URL, keeps the minted
+ * `state` in localStorage, and navigates away. The provider callback
+ * redirects back with a one-time `code` (or `error`); on mount this
+ * component redeems `code` + `state` for a token bundle and persists it.
+ *
+ * Cross-tab coordination (refresh mutex, storage-event sync) is deliberately
+ * omitted: the core's refresh grace window absorbs racing tabs, and a
+ * sibling tab self-corrects on its next refresh.
+ */
+export function AuthProvider({
```
Nicolapps, line 154: We already have a `AuthProvider` exported by the core. For the real implementation we might want to use something different.

examples/react-oauth/convex/auth.ts
```diff
@@ -0,0 +1,63 @@
...
+      authorizationEndpoint: "https://accounts.google.com/o/oauth2/v2/auth",
+      tokenEndpoint: "https://oauth2.googleapis.com/token",
+      issuer: "https://accounts.google.com",
+      scopes: ["openid", "email", "profile"],
+      pkce: true,
+      allowedRedirectOrigins: ["http://localhost:5173"],
+    }),
+    // GitHub is plain OAuth (no id_token), so identity comes from userinfo
+    // endpoints and a profile mapping. No `pkce`: GitHub OAuth apps silently
+    // ignore it.
```
Nicolapps, line 36: > No `pkce`: GitHub OAuth apps silently ignore it

This seems to be incorrect: GitHub seems to document PKCE support (https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/authorizing-oauth-apps)

examples/react-oauth/convex/auth.ts
Nicolapps: I think our APIs should make it harder not to use our 

There are many benefits for users to _not_ set up providers by themselves:
- They get typesafe profile data
- They won’t use providers in a way that’s less secure than possible (e.g. omitting PKCE, like this example is doing)
- They will not handle email data  incorrectly (e.g. forget to check if the email address is verified: OAuth itself has no standard way to indicate how trustworthy an email address is, and providers each have their own specific implementation).

We should still allow users to use custom OAuth providers, but the shape of the API should discourage that. For instance, the parameters could have a parameter like `provider: "google" | "github" | "microsoft" | … | "custom"`.

This would incentivize users to use the default implementation when one exists. In particular, I’m trying to avoid the scenario where a LLM already knows the configuration parameters from its training data, and doesn’t know about the `github()` wrapper.

What do you think?

Nicolapps: ^ I also realize that doing this will also be better if we later support a zero-config Shoo-like experience for using OAuth in dev.

I’m not sure how exactly the configuration parameters would look like. Would users omit the client ID/secret? Would they need to manually provide a `mode: "dev"` parameter? How would we make sure users can’t use this setup in prod? It’s probably worth thinking about this before the release of v2, even if we don’t support zero-config initially

erquhart:
Superseded by a restacked series starting at #431 - initial stack with just the backend and no client, rebuilt into self-contained layers per review feedback. Feedback from this PR's review is absorbed into the new stack, and will be included in one or two smaller stacks that tackle client and beyond.

PR #417, opened 2026-07-16
Split support for SPA and SSR sign-in and token handling

dowski:
When using SSR, client code will call through the SSR frontend server for all actions related to auth. The response will have the refresh token as an httpOnly cookie, not present in the token bundle returned to the client.

This change puts some of that groundwork in place.

----

### DIFF
--- a/packages/core/src/browser/sessionManager.ts
+++ b/packages/core/src/browser/sessionManager.ts
@@ -106,7 +106,7 @@ export class AuthClient {
    */
   readonly #refresh: () => Promise<TokenBundle | SlimTokenBundle | null>;
   /** Revoke the session on the server. */
-  readonly #signOutApi: () => Promise<void>;
+  readonly #signOutInternal: () => Promise<void>;
   readonly #storage: NamespacedStorage;
   readonly #verbose: boolean;
   readonly #lockKey: string;
@@ -141,7 +141,7 @@ export class AuthClient {
         if (refreshToken === null) return null;
         return authApi.refreshSession(refreshToken);
       };
-      this.#signOutApi = async () => {
+      this.#signOutInternal = async () => {
         const refreshToken = await this.#currentRefreshToken();
         if (refreshToken !== null) await authApi.signOut(refreshToken);
       };
@@ -150,7 +150,7 @@ export class AuthClient {
       // The refresh token is in an httpOnly cookie that the API reads
       // server-side, so it isn't passed directly.
       this.#refresh = () => authApi.refreshSession();
-      this.#signOutApi = () => authApi.signOut();
+      this.#signOutInternal = () => authApi.signOut();
     }
   }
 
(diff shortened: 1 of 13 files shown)


packages/core/src/browser/sessionManager.ts
```diff
@@ -275,6 +326,16 @@ export class AuthClient {
     this.#notify();
   }
 
+  /**
+   * Stores just the access token and notifies subscribers.
+   */
+  async #storeAccessOnly(session: SlimTokenBundle): Promise<void> {
+    this.#accessToken = session.accessToken;
+    await this.#storage.set(JWT_STORAGE_KEY, session.accessToken);
```
Nicolapps, line 354: If receiving a slim token bundle, should we clear the existing refresh token if there is one? I don’t think apps are likely to alternate between the two modes in practice, but it might be worth still doing it defensively

```suggestion
    this.#accessToken = session.accessToken;
    this.#refreshToken = null;
    await this.#storage.set(JWT_STORAGE_KEY, session.accessToken);
    await this.#storage.remove(REFRESH_TOKEN_STORAGE_KEY);
```

packages/core/src/react/index.tsx
```diff
@@ -111,10 +90,16 @@ export function ConvexAuthProvider({
     });
     return new AuthClient({
       authApi: {
+        // The browser holds the refresh token, so it is always present here; a
+        // `null` means there is no session to refresh, so short-circuit.
         refreshSession: (refreshToken) =>
-          httpClient.mutation(api.refreshSession, { refreshToken }),
+          refreshToken === null
+            ? Promise.resolve(null)
```
Nicolapps: Why do we skip the HTTP function call here? I would expect this code path to still do the HTTP request and return a slim token.

dowski: Thanks, this helped me see this wasn't very clear. For an SPA, when there's no refresh token a refresh should be a no-op. That's what this was trying to convey, but it's deep enough in the stack that it's confusing.

Now there are distinct modes for the `AuthClient` in `sessionManager.ts`: one for SPA and one for SSR, and the refresh and signout behavior vary on that.

packages/core/src/browser/sessionManager.ts
dowski: Done!

dowski:
* **#420** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/420?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#417** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/417?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/417?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `reboot`


<h2></h2>

Nicolapps (approved):

dowski:
### Merge activity

* **Jul 22, 6:14 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/get-convex/convex-auth/417).
* **Jul 22, 6:14 PM UTC**: @dowski merged this pull request with [Graphite](https://app.graphite.com/github/pr/get-convex/convex-auth/417).

PR #419, opened 2026-07-17
Add usePasswordSignIn and usePasswordSignUp hooks

dowski:
I think this is beneficial for a few reasons:

1. We take on some boilerplate that every app would need to replicate
2. `setSession` feels kind of low-level and not something app developers should be commonly reaching for
3. Apps can just handle the `pending` state rather than ensuring it's updated correctly
4. Having a common hook-based API for providers should make using each one feel somewhat familiar

----


erquhart:
Looks good 👍

Nicolapps (approved):

PR #420, opened 2026-07-21
Add framework-agnostic SSR cookie and auth handlers

dowski:
Add framework-agnostic SSR cookie and auth handlers

Uses the public `cookie` package (which would wind up as a transitive dependency once Next.js is integrated). Ensures that the refresh and access tokens are properly encoded in cookies (for use on the server) as well as returning a `SlimTokenBundle`for client use of the access token.

Exposes common configuration/setup for the handlers required to implement SSR with Convex Auth. The refresh, sign-out and sign-in handlers are standard `Request`/`Response` functions that should be able to be adapted to any SSR framework.

### DIFF
--- a/packages/core/src/server/httpCookies.ts
+++ b/packages/core/src/server/httpCookies.ts
@@ -72,11 +72,12 @@ export function httpCookies(request: Request): HttpCookies {
       overlay.set(name, value);
       pending.push(serializeCookie(name, value, options));
     },
-    delete(name) {
+    delete(name, options) {
       overlay.set(name, null);
       pending.push(
         serializeCookie(name, "", {
-          path: "/",
+          path: options?.path ?? "/",
+          domain: options?.domain,
           maxAge: 0,
           expires: new Date(0),
         }),
--- a/packages/core/src/server/httpCookies.ts
+++ b/packages/core/src/server/httpCookies.ts
@@ -1,18 +1,3 @@
-/**
- * A framework-agnostic {@link CookieStore} over a WHATWG `Request`/`Response`.
- *
- * The shared {@link ServerAuthSession} reads and writes cookies through a
- * {@link CookieStore}. A framework that hands off raw WHATWG requests (a route
- * handler that is `(Request) => Response` — Next.js App Router, Remix, Hono, …)
- * can use {@link httpCookies} to adapt them: it parses the incoming `Cookie`
- * header for reads and buffers writes as `Set-Cookie` headers to apply to the
- * outgoing response.
- *
- * Nothing here depends on React, Next.js, or Convex.
- *
- * @module
- */
-
 import { parse, serialize } from "cookie";
 import { CookieOptions, CookieStore } from "./cookies";
 
@@ -54,7 +39,11 @@ export interface HttpCookies extends CookieStore {
 }
 
 /**
- * Adapt an incoming WHATWG {@link Request}'s cookies to a {@link CookieStore}.
+ * Adapt an incoming {@link Request}'s cookies to a {@link CookieStore}.
+ *
+ * The returned {@link HttpCookies} object allows applying changes to the
+ * cookies via the {@link HttpCookies#applyTo} function. Pass it {@link
+ * Response} `Headers` and it will write the cookie values.
  *
  * ```ts
  * const cookies = httpCookies(request);
--- a/packages/core/src/server/httpCookies.ts
+++ b/packages/core/src/server/httpCookies.ts
@@ -13,12 +13,14 @@
  * @module
  */
 
+import { parse, serialize } from "cookie";
 import { CookieOptions, CookieStore } from "./cookies";
-
-const SAME_SITE = { lax: "Lax", strict: "Strict", none: "None" } as const;
 
 /**
  * Serialize a cookie into a `Set-Cookie` header value.
+ *
+ * Thin adapter over the `cookie` package's `serialize` that maps our
+ * {@link CookieOptions} onto it.
  *
  * Deletion is expressed by the caller as an empty value with `maxAge: 0` (and a
  * past `expires`); see {@link httpCookies}'s `delete`.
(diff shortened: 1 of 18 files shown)


dowski:
* **#435** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/435?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#428** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/428?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#427** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/427?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#420** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/420?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/420?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#425** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/425?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `reboot`


<h2></h2>

erquhart:
I hit "review with Graphite" and it started doing some kind of AI review I think? I did not intend that. Still reviewing.

packages/core/src/server/httpCookies.ts
```diff
@@ -0,0 +1,89 @@
...
+  return {
+    get(name) {
+      if (overlay.has(name)) return overlay.get(name) ?? undefined;
+      return jar[name];
+    },
+    set(name, value, options) {
+      overlay.set(name, value);
+      pending.push(serializeCookie(name, value, options));
+    },
+    delete(name) {
```
erquhart: Cookies set with a custom domain/path wouldn't get deleted by just a name

packages/core/src/server/setup.ts
```diff
@@ -0,0 +1,107 @@
...
+ *   refreshSession: api.auth.refreshSession,
+ *   signOut: api.auth.signOut,
+ *   cookieOptions: { secure: process.env.NODE_ENV === "production" },
+ * });
+ *
+ * // in a route file:
+ * export const POST = auth.signInHandler(anonymous(api.auth.signInAnonymous));
+ * ```
+ */
+export function setupConvexAuthServer(config: ConvexAuthServerConfig) {
```
erquhart, line 102: Unless this is on hold for a future hardening round, we can add a CSRF guard here to enforce origin, method, and check Sec-Fetch-Site header https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#how-to-treat-fetch-metadata-headers-on-the-server-side

erquhart (commented):
These might be worth passing from my agent to yours:

- writeAuthCookies`merges as`{ ...invariants, ...options }`, so a caller can silently pass `httpOnly: false`or`sameSite: "none"`and break the "refresh token never reaches client JS" invariant. The shape is pre-existing, but #420's`setupConvexAuthServer`makes`cookieOptions`the single public knob, so it's now a one-liner for an integrator to do this. Omitting`httpOnly`/`sameSite`from`AuthCookieOptions` (or spreading invariants last) would lock it down.
- `refreshHandler` lets backend/network errors propagate as thrown exceptions (a 500 with no cookies applied) while `signOutHandler` catches everything. Propagating is arguably right (don't treat a transient failure as signed-out), but the asymmetry looks accidental; worth a comment if intentional.
- Question: packages/core/src/components/anonymous/server.ts lives inside the component directory, so it gets pushed to every consuming deployment as a component module, bundling `cookie` and `ConvexHttpClient`. It follows the existing react.tsx precedent and codegen/CI accepted it, but SSR-only code in the pushed component bundle is dead weight; was keeping it outside the component dir considered?
- Nits: `"cookie": "1.1.1"` is exact-pinned while sibling deps use carets; and a refresh request with a missing refresh cookie but present JWT cookie returns `{tokens: null}` without clearing the stale JWT cookie (bounded by the 1-minute access TTL).

packages/core/src/server/setup.ts
dowski: Thanks! Actually used Origin checking since it predates Sec-Fetch-Site. WDYT?

packages/core/src/server/httpCookies.ts
dowski: Good catch! Should be fixed up now.

dowski:
Thanks, I thought I replied here but I guess I lost it somehow.  
  
I updated the fix that cookie issue - thanks! The handler error difference is already documented in the sign out code. The other two seem okay to address later.

packages/core/src/server/origin.ts
```diff
@@ -0,0 +1,62 @@
...
+ * Browsers attach an `Origin` header to every cross-site POST, so the rule
+ * is: a request whose `Origin` matches neither the `Host` header it arrived
+ * at nor a configured allowed origin is refused, before any Convex call runs
+ * or cookie is written.
+ *
+ * @module
+ */
+
+/** An origin's host, tolerating a bare host (`"app.example.com"`) as input. */
+function hostOf(origin: string): string {
```
erquhart, line 26: Would fail with ports, eg., localhost:3000

```suggestion
function hostOf(origin: string): string {
  // A bare host with a port parses as a URL *scheme* ("localhost:3000" →
  // scheme "localhost:", host ""), so only URL-parse full origins.
  if (!origin.includes("://")) return origin;
```

erquhart (approved):
One suggestion to fix, pre-approving pending that.

packages/core/src/server/origin.ts
dowski: Good catch, thanks!

dowski:
### Merge activity

* **Jul 30, 2:40 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/get-convex/convex-auth/420).
* **Jul 30, 2:40 PM UTC**: @dowski merged this pull request with [Graphite](https://app.graphite.com/github/pr/get-convex/convex-auth/420).

PR #423, opened 2026-07-22
fix(auth): harden concurrency, connections, and lifecycle safety

robelest:
## What changed

This is a broad pre-0.1 hardening pass across the auth component and its public facades.

- Makes API-key verification/rotation, OAuth code redemption and refresh issuance, device approval, passkey counters, TOTP verifier consumption, account linking, and SCIM identity provisioning safe under concurrent requests.
- Hardens SAML replay/certificate handling, outbound SSRF filtering, client auth-state recovery, OAuth scope/email semantics, CORS/cookie behavior, and sensitive record projection.
- Makes large group hierarchy cascades durable and resumable, fixes membership ordering, and bounds revoked OAuth-client retention.
- Replaces the impossible legacy webhook hash “re-encryption” path with a safe disable/rotate migration and prevents delivery without encrypted signing material.
- Aligns the public connection-admin surface with the project lexicon and regenerates the component bindings.
- Adds regression coverage for the concurrency, migration, lifecycle, and contract failures found during two independent deep-review passes.

## Why

The branch originally addressed a large audit backlog, but a second review found several check-then-write races and migration assumptions that could still produce duplicate sessions, stale authorization, orphaned SCIM users, unrecoverable one-time credentials, partial group cascades, or unusable webhook secrets. The follow-up commit moves the winning transitions into Convex mutations, makes post-commit audit delivery best-effort where a returned secret cannot be recovered, and documents the required operator rollout.

## Breaking changes

- Connection admin methods are renamed to the canonical verbs documented in `packages/auth/MIGRATION-vNext.md` (`set` → `upsert`, webhook `disable` → `revoke`, request `add` → `mount`).
- OAuth email linking now requires an explicit verified-email signal.
- Structured user names add `firstName` / `lastName` alongside the formatted `name` field so SCIM mappings no longer write undeclared User keys.

## Upgrade / rollout

After deployment, run:

```sh
vp exec convex run auth/migrations:runDisableLegacyWebhookEndpoints '{}'
vp exec convex run auth/migrations:runBackfillOAuthClientRevokedAt '{}'
```

The webhook migration disables hash-only legacy endpoints because their original HMAC secrets cannot be recovered. Rotate each subscriber secret and explicitly reactivate the endpoint (or replace it). The OAuth-client migration conservatively starts the 90-day retention clock for legacy revoked registrations.

See `packages/auth/MIGRATION-vNext.md` and the migration reference for the complete vNext migration sequence.

## Validation

- `vp install`
- Convex component and app codegen
- `vp run cache:build:auth`
- `vp run typecheck`
- `vp lint`
- `vp fmt --check .`
- `vp run cache:test:unit` — 66 files, 508 tests passed
- `vp run check:docs` — 0 errors, 0 warnings
- `vp run '@robelest/convex-auth#check:packaging'` — attw/publint passed

Docker-backed interop tests were attempted but could not start because the local Docker/OrbStack daemon is unavailable. The non-Docker test suite completed successfully before that infrastructure failure.


robelest:
Closing this draft because it targeted the upstream repository rather than the feature branch's actual base, robelest/convex-auth:main. A correctly based draft will replace it.

PR #425, opened 2026-07-23
Add framework-agnostic server auth token verification

dowski:
This will be used by SSR code to validate authentication status in middleware and server components.

PR #422, opened 2026-07-21
Add Next.js SSR example (examples/nextjs)

dowski:
An App Router example of server-side sign-in with Convex Auth. Sign-in,
refresh, and sign-out are mounted as route handlers under app/auth/ from the
framework-agnostic handlers:

- app/auth/signin/anonymous -> anonymousSignInHandler
- app/auth/refresh          -> refreshHandler
- app/auth/signout          -> signOutHandler

Uses the setup/configuration functionality exposed for Next.js apps.

PR #421, opened 2026-07-21
Add Next.js App Router bindings

dowski:
Provides middleware, auth helpers and client and server components that assist in setting up a Next.js app with Convex Auth.

PR #427, opened 2026-07-27
Modifications to the SSR core to support other sign in patterns

dowski:
The major change is that all sign-in/sign-up code can be written as plain React hooks and for SSR those will be proxied through the SSR host to the Convex backend. The client-side code will use a ConvexHttpClient for those requests which is what lets them share the hooks. The SSR side code will ensure that the refresh tokens get set properly in httpOnly cookies and are removed from response bodies.  
  
The immediate use case is username/password but this will generalize to other sign-in/sign-up code as well.


dowski:
* **#435** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/435?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#428** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/428?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#427** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/427?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/427?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#420** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/420?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#425** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/425?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `reboot`


<h2></h2>

packages/core/src/server/handlers.ts
```diff
@@ -117,27 +121,57 @@ export function signOutHandler(config: SignOutHandlerConfig): RequestHandler {
...
+  | {
+      /** The minted session tokens. */
+      tokens: TokenBundle;
+    }
+  | {
+      /** `null`: sign-in failed. */
+      tokens: null;
+      /** Provider-specific failure detail, echoed to the client verbatim. */
+      userError?: unknown;
+    };
```
Nicolapps: nit: What is the reason why we’re not using the same format as the current functions? I think it’s a little bit more explicit, and it avoid the need for the `toOutcome` conversion.

```suggestion
  | {
      success: true,
      /** The minted session tokens. */
      tokens: TokenBundle;
    }
  | {
      success: false,
      /** Provider-specific failure detail, echoed to the client verbatim. */
      userError: unknown;
    };
```

packages/core/src/server/handlers.ts
```diff
@@ -117,27 +121,57 @@ export function signOutHandler(config: SignOutHandlerConfig): RequestHandler {
...
+  if (outcome.tokens !== null) {
+    await writeAuthCookies(cookies, outcome.tokens, cookieOptions);
+    res = Response.json({ tokens: makeSlimBundle(outcome.tokens) });
+  } else {
+    const { userError } = outcome;
+    res = Response.json(
+      { tokens: null, ...(userError !== undefined && { userError }) },
+      { status: 401 },
+    );
+  }
```
Nicolapps: I think it could be useful to define a `SlimSignInOutcome` type, and make sure the response here matches it

(probably, this type could also contain a `success` field)

```suggestion
  if (outcome.tokens !== null) {
    await writeAuthCookies(cookies, outcome.tokens, cookieOptions);
    res = Response.json({ tokens: makeSlimBundle(outcome.tokens) } satisfies SlimSignInOutcome);
  } else {
    const { userError } = outcome;
    res = Response.json(
      { tokens: null, ...(userError !== undefined && { userError }) } satisfies SlimSignInOutcome,
      { status: 401 },
    );
  }
```

packages/core/src/server/handlers.ts
```diff
@@ -69,9 +72,10 @@ export function refreshHandler(config: RefreshHandlerConfig): RequestHandler {
       cookieOptions: config.cookieOptions,
     });
     const bundle = await session.refresh();
-    const res = Response.json({
-      tokens: bundle === null ? null : makeSlimBundle(bundle),
-    });
+    const res =
+      bundle === null
+        ? Response.json({ tokens: null }, { status: 401 })
+        : Response.json({ tokens: makeSlimBundle(bundle) });
```
Nicolapps: Similar comment here: it would be useful to verify the result format is right before JSON-encoding it

packages/core/src/components/password/server.ts
```diff
@@ -0,0 +1,87 @@
+/**
```
Nicolapps: Do we need specific server code for login flows like passwords? Now that we’re supporting generic functions, can’t we avoid needing to do the argument parsing on the server?

Nicolapps: (I think it's still helpful to have the proxy allowlist a list of server endpoints that it is allowed to contact — but I don’t think it should do the parsing specifically for `{ username, password }`)

dowski: Thanks for the question here. I think the work later in this branch only partially solves what you are asking about.  
  
I think what you're identifying (having to duplicate SSR handler code for every sign-in method) is worth fixing. I'm taking a look at that and I'll get back here with updated code.

packages/core/src/server/handlers.ts
dowski: Removed!

packages/core/src/server/handlers.ts
dowski: Removed `SignInOutcome`

packages/core/src/server/handlers.ts
dowski: I might not fully understand. We're handling tokens that were returned by a Convex backend here.

dowski (commented):
Okay, I think this is in better shape now. Thanks for the review suggestions! I actually wound up dropping the PR that was previously on the top of the stack b/c this one subsumed it.

packages/core/src/react/client.tsx
```diff
@@ -12,6 +17,55 @@ import {
...
+ *  - SPA: called against the deployment, returning the full
+ *    {@link TokenBundle} for client JS to persist.
+ *  - SSR: called through the auth proxy on the SSR host, which moves the
+ *    refresh token into an httpOnly cookie and returns an access-only
+ *    {@link SlimTokenBundle}.
+ *
+ * A provider never sees which model it is running under. See
+ * {@link ClientView} for why one declared type is honest for both.
+ */
+export interface AuthRunner {
```
Nicolapps: Do you think that it would make sense to use AuthRunner for session refresh too?

If not, we should maybe consider giving it a name that's a little bit more explicit (e.g. `AuthSessionCreationClient`?)

packages/core/src/server/setup.ts
```diff
@@ -76,6 +27,13 @@ export interface ConvexAuthServerConfig {
   refreshSession: RefreshSessionFn;
   /** The app's `signOut` mutation reference. */
   signOut: SignOutFn;
+  /**
+   * The provider sign-in functions to expose through the auth proxy.
```
Nicolapps, line 31:
```suggestion
   * The provider sign-in functions to expose through the auth proxy.
   * This should include all the Convex backend functions in your app that can issue auth sessions (e.g. account creation, log in with existing account…).
```

packages/core/src/server/handlers.ts
Nicolapps: This comment is suggesting that we should have a TypeScript type that defines the expected output format of the refresh route (e.g. `type … = { tokens: … }`, and that we make sure that whatever we pass to `Response.json` matches that type (e.g. `Response.json({ … } satisfies …)`)

examples/nextjs/app/auth/proxy/route.ts
```diff
@@ -0,0 +1,10 @@
+import { auth } from "@/src/lib/serverAuth";
+
+// The auth proxy, mounted once. It serves the `ConvexHttpClient` endpoints
+// (/auth/api/mutation, /auth/api/action) for every provider listed in
+// `signIn` in serverAuth.ts, forwarding each call to the deployment and moving
+// the minted refresh token into an httpOnly cookie on the way back.
+//
+// Adding an auth method means adding its function to that allowlist. There is
+// no per-method route or client code.
+export const POST = auth.proxyHandler;
```
Nicolapps: We can avoid the catch-all route, per our live discussion

dowski: Thanks, done!

packages/core/src/react/client.tsx
dowski: Renamed it, thanks.

packages/core/src/server/handlers.ts
dowski: Got it, done.

examples/nextjs/app/auth/proxy/route.ts
```diff
@@ -0,0 +1,10 @@
+import { auth } from "@/src/lib/serverAuth";
+
+// The auth proxy, mounted once. It serves the `ConvexHttpClient` endpoints
+// (/auth/proxy?path=/api/mutation, ?path=/api/action) for every provider listed
+// in `signIn` in serverAuth.ts, forwarding each call to the deployment and
+// moving the minted refresh token into an httpOnly cookie on the way back.
```
Nicolapps: I think it’s worth simplifying this comment to avoid talking about the implementation details (ConvexHttpClient/`?path=`), since we will probably include this file as part of our docs.

```suggestion
// The SSR auth proxy. It proxies the auth sign-in routes set up in serverAuth.ts, forwarding each call to the deployment. Instead of sending all resulting tokens as part of the response body (as calling the Convex function directly would do), the proxy moves the minted refresh token into an httpOnly cookie on the way back.
```

examples/nextjs/README.md
```diff
@@ -8,17 +8,24 @@ access token.
 
 - **Sign-in / refresh / sign-out run on the server** as framework-agnostic
   `(Request) => Response` handlers, mounted under `app/auth/`:
-  - `app/auth/signin/anonymous/route.ts` → `anonymousSignInHandler`
+  - `app/auth/proxy/route.ts` → `proxyHandler`, the auth proxy for every
```
Nicolapps: I wonder if this name could mislead users to think that it’s a [Next.js proxy route](https://nextjs.org/docs/app/getting-started/proxy) (`proxy.ts`). What do you think about renaming the route to something like `/api/auth/signIn`?

Nicolapps (approved):

examples/nextjs/README.md
dowski: Good point - done.

dowski:
### Merge activity

* **Aug 12, 10:28 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/get-convex/convex-auth/427).
* **Aug 12, 10:28 PM UTC**: [Graphite](https://app.graphite.com/github/pr/get-convex/convex-auth/427) rebased this pull request as part of a merge.
* **Aug 12, 10:31 PM UTC**: @dowski merged this pull request with [Graphite](https://app.graphite.com/github/pr/get-convex/convex-auth/427).

PR #431, opened 2026-07-28
Add OAuth component sign-in leg: authorization requests, startSignIn, config validation

erquhart:
First of a restacked series superseding #411.

- Add the OAuth provider component with its sign-in leg: `startSignIn` creates a short-lived authorization request for a validated redirect target
- Validate provider config at setup
- Seed KNOWN_ISSUES.md with the limitations introduced here


erquhart:
* **#433** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/433?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#432** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/432?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#431** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/431?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/431?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `reboot`


<h2></h2>

packages/core/src/oauth/component/convex.config.ts
```diff
@@ -0,0 +1,36 @@
...
+ *     GITHUB_CLIENT_SECRET: app.env.AUTH_GITHUB_CLIENT_SECRET,
+ *   },
+ * });
+ * ```
+ *
+ * The credential bindings are what let each provider's pair reach the
+ * component's callback (for the token exchange) without ever being stored in a
+ * table. Every pair is optional: an app binds only the providers it uses, and a
+ * `provider(...)` wired without its matching binding fails at the first sign-in
+ * (see `credentials.ts`) rather than at push, since the mount bindings aren't
```
erquhart: Note: this isn't actually a good thing, it's just what we have. Validating at push time would require some way to dynamically validate or for each oauth provider to use a separate component. I suspect the planned `init` work would help here if components are able to participate.

packages/core/src/oauth/component/provider.ts
```diff
@@ -0,0 +1,101 @@
...
+const AUTHORIZATION_REQUEST_TTL_MS = 10 * 60 * 1000;
+
+/**
+ * Record an in-flight authorization request. Called by the app-side
+ * `signIn` before it redirects the user to the provider; the provider
+ * callback later claims the request by state hash.
+ *
+ * Exchange config the callback needs (`callbackUrl`, `tokenEndpoint`,
+ * `userInfoEndpoints`) is snapshotted onto the row here because the
+ * component can't see app-side config or system env vars — a typed-env
```
erquhart: Support for system env vars to be visible for all functions everywhere is merged but not released, but even then relying on it sets a really high convex package minimum version. For now I've left this as is, can update to use system env vars when that change is released.

packages/core/src/oauth/component/setup.ts
```diff
@@ -0,0 +1,275 @@
...
+   * `https://accounts.google.com`. Present for OIDC providers; absent for
+   * plain-OAuth providers (GitHub), where identity comes from userinfo.
+   */
+  issuer?: string;
+  /**
+   * Endpoints the callback fetches with the access token (GET with a bearer
+   * token), keyed by the name the profile mapping reads each response under.
+   * Present for providers whose identity comes from userinfo (GitHub).
+   */
+  userInfoEndpoints?: Record<string, string>;
```
erquhart, line 35: This is a flexible interface that should serve well for the varying calls any provider might use to gather user info. Each provider that has user info endpoints can configure this and then map the responses onto a single profile, so for the user they just get a single profile and don't see this part.

packages/core/src/oauth/component/constants.ts
erquhart, line 1: Note: the `src/oauth/component` layout (instead of `src/components/oauth`) is a reversible recommendation - structure each component as we would if it were its own package, which some of them probably will eventually be. A component typically has its own component directory alongside non-component code like react code. The oauth component here follows that pattern.

packages/core/src/oauth/component/convex.config.ts
```diff
@@ -0,0 +1,36 @@
...
+ * visible to the app-side setup that would otherwise catch it.
+ */
+const component = defineComponent("oauth", {
+  env: {
+    GOOGLE_CLIENT_ID: v.optional(v.string()),
+    GOOGLE_CLIENT_SECRET: v.optional(v.string()),
+    GITHUB_CLIENT_ID: v.optional(v.string()),
+    GITHUB_CLIENT_SECRET: v.optional(v.string()),
+  },
+});
```
dowski, line 33: Can the component take the name of the provider as an argument? Then you mount multiple instances of the component, one for each provider you want to support in your app. That would allow one set of non-optional client-secret/client-id env vars per configured provider.

packages/core/src/oauth/component/credentials.ts
```diff
@@ -0,0 +1,55 @@
...
+  const bindings = CREDENTIAL_BINDINGS[provider];
+  const clientId = env[bindings.id];
+  const clientSecret = env[bindings.secret];
+  if (!clientId || !clientSecret) {
+    throw new Error(
+      `Missing credentials for OAuth provider "${provider}". Bind ${bindings.id} and ${bindings.secret} on the oauth mount in convex.config.ts.`,
+    );
+  }
+  return { clientId, clientSecret };
+}
```
dowski: I think if each provider had its own instance of the component (as described in my comment earlier) then this all goes away.

packages/core/src/oauth/component/oauth.test.ts
```diff
@@ -0,0 +1,130 @@
...
+      expect(requests[0].expiresAt).toBeGreaterThan(Date.now());
+    });
+  });
+
+  test("createAuthorizationRequest throws for an unsupported provider", async () => {
+    const t = setup();
+    await expect(
+      t.mutation(api.provider.createAuthorizationRequest, {
+        ...requestArgs,
+        provider: "myspace",
```
dowski: lol

packages/core/src/oauth/component/provider.ts
dowski: Sounds good - can you add a TODO?

packages/core/src/oauth/component/provider.ts
```diff
@@ -0,0 +1,101 @@
...
+    return { clientId };
+  },
+});
+
+/**
+ * Claim an authorization request by state hash: find, delete, and return it
+ * in one transaction, so a replayed or raced callback finds nothing. An
+ * expired row is also deleted, but only its `redirectTo` is returned so the
+ * callback can send the user back to the app instead of stranding them on
+ * an error page.
```
dowski: Can you preface this with a note that it's a part of the second half of the OAuth authorization process? The comment describes the "what" so far, but not the "why".

dowski: Maybe note as well that this is expected to be called by the component's HTTP callback route.

packages/core/src/oauth/component/schema.ts
```diff
@@ -0,0 +1,51 @@
...
+     * byte-identical value, as OAuth requires.
+     */
+    callbackUrl: v.string(),
+    /**
+     * PKCE code verifier, present only when the provider config enables PKCE.
+     * Stored raw because it must be sent to the provider at code exchange.
+     */
+    codeVerifier: v.optional(v.string()),
+    /**
+     * The provider's token endpoint, copied from app-side config at
```
dowski: This is a URL, right? Maybe include that in the name, just like how `callbackUrl` has it.  
  
Or at least note it in the doc.

packages/core/src/oauth/component/schema.ts
```diff
@@ -0,0 +1,51 @@
...
+     * sign-in: the callback runs inside the component, which can't see app
+     * config, so non-secret exchange config is stored on the row.
+     */
+    tokenEndpoint: v.string(),
+    /**
+     * Profile endpoints to fetch with the access token after the exchange,
+     * keyed by the name the app's `profile` mapping receives each response
+     * under. Copied from app-side config like `tokenEndpoint`.
+     */
+    userInfoEndpoints: v.optional(v.record(v.string(), v.string())),
```
dowski, line 30: Similar thought re: URL here.

packages/core/src/oauth/component/schema.ts
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * Expected `iss` of the provider's id_tokens, copied from app-side
```
dowski: Note that `iss` is part of the OIDC spec.

packages/core/src/oauth/component/setup.ts
```diff
@@ -0,0 +1,276 @@
...
+
+/**
+ * The library-owned config for one IdP: everything that describes how to
+ * talk to the provider. Shipped by each supported provider's catalog module,
+ * not by the app. These values are constants, so they aren't re-validated at
+ * setup the way user input is.
+ */
+export type OauthCatalog = {
+  /** The provider's authorization endpoint, e.g. Google's `https://accounts.google.com/o/oauth2/v2/auth`. */
+  authorizationEndpoint: string;
```
dowski, line 16: Same nit re: naming w/out URL, and relates to `userInfoEndpoints` below too.

packages/core/src/oauth/component/setup.ts
```diff
@@ -0,0 +1,276 @@
...
+ * be re-exported side by side without renaming.
+ *
+ * The component is mounted once in `convex.config.ts` with each provider's
+ * `<PROVIDER>_CLIENT_ID`/`_SECRET` bindings; register
+ * `<site-url><httpPrefix>/<provider>/callback` as the redirect URI with each
+ * provider.
+ */
+export function Oauth<const N extends string>(name: N, catalog: OauthCatalog) {
+  return defineProvider({
+    name,
```
dowski: nit: maybe call this `oauthProviderName`. It reads more clearly at usage sites below.

dowski (changes requested):

packages/core/src/oauth/component/convex.config.ts
erquhart: Yeah this was mean to avoid having n tables/registered components when you have a lot of oauth providers, but the ergonomics are better with a per-provider registration. Rewriting.

packages/core/src/oauth/component/schema.ts
erquhart: This and the other related ones I'd push back on - they're named from the oidc spec (`authorization_endpoint`, `token_endpoint`, `userinfo_endpoint`). Discovery metadata names. Diverging would feel like a drawback, and `endpoint` does sort of inherently carry the url concept. Let me know if you want to do this anyway.

dowski: That makes sense - thanks!

packages/core/src/oauth/component/provider.ts
erquhart: Support is live now so this is actually resolved, no todo needed

packages/core/src/oauth/component/setup.ts
erquhart: It's in the context of the Oauth component so I did `providerName` - say the word and I'll do literal `oauthProviderName`

packages/core/src/oauth/component/crypto.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Crypto helpers for the OAuth provider. Used by the app-side recipe to mint
+ * flow secrets (the state, the PKCE verifier and its challenge).
+ */
```
dowski: I'm not sure what happens, but I wonder if this doc comment gets attached to the function below rather than being tied to the module. I've seen `@module` syntax elsewhere, maybe that's needed here?

packages/core/src/oauth/component/provider.ts
```diff
@@ -0,0 +1,108 @@
...
+
+/** How long an authorization request stays claimable by the callback. */
+const AUTHORIZATION_REQUEST_TTL_MS = 10 * 60 * 1000; // 10m
+
+/**
+ * Record an in-flight authorization request. Called by the app-side
+ * `signIn` before it redirects the user to the provider; the provider
+ * callback later claims the request by state hash.
+ *
+ * Returns the mount's `CLIENT_ID` and the callback URL, which the caller
```
dowski: nitty-nit: maybe this should say "provider's `CLIENT_ID`"? I think that makes more sense in context (a `providerName` gets passed in).

packages/core/src/oauth/component/provider.ts
```diff
@@ -0,0 +1,108 @@
...
+  returns: v.object({
+    clientId: v.string(),
+    callbackUrl: v.string(),
+  }),
+  handler: async (ctx, args) => {
+    // System env vars are only visible inside components on backends with
+    // get-convex/convex-backend@64c163a (July 2026); cloud always has it,
+    // self-hosted may not.
+    // TODO: remove this check when no longer needed.
+    const siteUrl = process.env.CONVEX_SITE_URL;
```
dowski, line 35: I think Ian added the system env vars to the global `env` object. You can probably use those and remove any runtime checking (they'll only statically be present in more recent versions of the Convex package/API).  
  
I guess someone could run a later Convex package against an older server binary though? I'm going to ask about that in chat I think.

packages/core/src/oauth/component/provider.ts
```diff
@@ -0,0 +1,108 @@
...
+    if (siteUrl === undefined) {
+      throw new Error(
+        "CONVEX_SITE_URL is not visible inside the oauth component. " +
+          "This requires a Convex backend with system env vars in component " +
+          "functions (get-convex/convex-backend@64c163a, July 2026).",
+      );
+    }
+    const callbackUrl = `${siteUrl}${CALLBACK_PATH}`;
+    await ctx.db.insert("authorizationRequests", {
+      ...args,
```
dowski, line 45: Should we do some validation on the args?  
  
Like ensure the `providerName` is valid, that the `redirectTo` is allowed, etc?  
  
Maybe `providerName` would be hard to assert due to where it gets defined and where this code executes?  
  
Also I might just be getting thrown off by seeing `mutation` here, but this is a component so this isn't really public. You can probably ignores this comment :upside_down_face:

packages/core/src/oauth/component/setup.ts
```diff
@@ -0,0 +1,201 @@
...
+  allowedRedirectOrigins: string[];
+  /**
+   * Override the provider's default scopes (e.g. to request extra scopes).
+   */
+  scopes?: string[];
+  /**
+   * Extra query params for the authorization URL, e.g. Google's
+   * `{ access_type: "offline" }`. Must not include protocol params (listed below).
+   */
+  extraAuthorizationParams?: Record<string, string>;
```
dowski: Maybe we should leave these out at this point, since we're just focused on using OAuth for point-in-time authentication, not additional access or for accessing extra info about the user?

packages/core/src/oauth/component/setup.ts
```diff
@@ -0,0 +1,201 @@
...
+  "code_challenge_method",
+];
+
+/**
+ * Set up OAuth sign-in against a single upstream identity provider.
+ *
+ * This is the first leg of the flow: `startSignIn` validates, mints `state`,
+ * stores an authorization request record in the component, and returns the provider
+ * authorization URL for the client to navigate to plus the state it must hold
+ * onto. The provider later redirects back to the mount's HTTP callback
```
dowski: Here I am complaining about mount again.

packages/core/src/oauth/component/setup.ts
```diff
@@ -0,0 +1,201 @@
...
+ */
+export function setupOauth(
+  providerName: string,
+  catalog: OauthCatalog,
+  helpers: ProviderHelpers,
+  options: OauthProviderOptions,
+) {
+  // Validate the app-supplied options up front so mistakes fail at deploy
+  // time, not on the first sign-in.
+  const allowedOrigins = options.allowedRedirectOrigins.map((allowed) => {
```
dowski, line 87: Should this be called `origin` or `rawOrigin` instead of `allowed`? It's not known whether it is allowed or not at this point.

packages/core/src/oauth/component/setup.ts
dowski: Will any of our packaged providers other than GitHub use this? And similarly, PKCE won't be used by all.  
  
We don't need to change direction right now, but it might be worth considering dedicated components per provider, where the schema/args can be tight and enforce requirements.

dowski (approved):

packages/core/src/oauth/component/setup.ts
erquhart: The reason I shied away from per-provider components was repo bulk and potentially bundle size, although that's probably something we can avoid if we're intentional. But yeah, will keep in mind. This pattern is used by other oauth providers, so for that part specifically this is meant to work as a reusable interface.

packages/core/src/oauth/component/setup.ts
erquhart: They're all allowed origins, this is just validating the list. `startSignIn` args.redirectTo is the url that might not be allowed, so we use this list as a check.

packages/core/src/oauth/component/provider.ts
erquhart: Held off for now because that hasn't actually been released, but we can circle back and add

packages/core/src/oauth/component/provider.ts
erquhart: yeah this is only reachable from our own code and the checks are in the caller

packages/core/src/oauth/component/setup.ts
erquhart: yeah no point in including them without an immediate use case, will drop

packages/core/src/oauth/component/crypto.ts
erquhart: only way to fix is to use simple // comments it seems, will do that

PR #432, opened 2026-07-28
Add OAuth callback leg: code exchange, id_token validation, one-time tickets

erquhart:
- Add the provider callback route: verify state, exchange the code, validate the provider's response, and redirect back to the app with an encrypted one-time ticket

### DIFF
--- a/KNOWN_ISSUES.md
+++ b/KNOWN_ISSUES.md
@@ -8,9 +8,9 @@ Probably all public routes need rate limiting of some sort.
 
 ## OAuth component rows are never cleaned up, and `startSignIn` is unauthenticated
 
-Expired authorization requests are only deleted when their state is later
-presented (`packages/core/src/oauth/component/provider.ts`), so abandoned
-flows accumulate forever.
+Expired authorization requests and tickets are only deleted when their
+secret is later presented (`packages/core/src/oauth/component/provider.ts`),
+so abandoned flows accumulate forever.
 
 ## OAuth sign-in requires a backend with system env vars in components
 
--- a/KNOWN_ISSUES.md
+++ b/KNOWN_ISSUES.md
@@ -29,3 +29,15 @@ No action item, just here for awareness.
 (`myapp://`, `exp://`) have a `"null"` origin under the URL standard and are
 rejected at setup (`packages/core/src/oauth/component/setup.ts`), so React
 Native apps must return via https universal links / app links.
+
+## Apple sign-in isn't supported yet
+
+Two gaps. The callback route only accepts GET redirects, and Apple POSTs the
+callback (`response_mode=form_post`) whenever name/email scopes are
+requested. And Apple has no static client secret: it requires a short-lived
+ES256 client-secret JWT, signed with a registered key and rotated, where the
+catalogs assume a static `CLIENT_SECRET` binding.
+
+Fix direction, wanted for launch: accept POST on the callback route
+(`packages/core/src/oauth/component/http.ts`), and add a signed-secret
+mechanism to the catalog config (`packages/core/src/oauth/component/setup.ts`).
--- a/KNOWN_ISSUES.md
+++ b/KNOWN_ISSUES.md
@@ -38,6 +38,6 @@ requested. And Apple has no static client secret: it requires a short-lived
 ES256 client-secret JWT, signed with a registered key and rotated, where the
 catalogs assume a static `CLIENT_SECRET` binding.
 
-Fix direction, wanted for launch: accept POST on the callback route
+Potential fix direction: accept POST on the callback route
 (`packages/core/src/oauth/component/http.ts`), and add a signed-secret
 mechanism to the catalog config (`packages/core/src/oauth/component/setup.ts`).
--- a/packages/core/src/lib/oauthParams.ts
+++ b/packages/core/src/lib/oauthParams.ts
@@ -1,0 +1,23 @@
+/**
+ * The query params the OAuth component's callback appends when it redirects the
+ * browser back to the app, and that the browser client reads on mount.
+ *
+ * They are deliberately namespaced rather than the bare `code`/`error` an OAuth
+ * callback would normally use. The client's callback handler is registered
+ * unconditionally (every app that uses Convex Auth runs it on mount), so bare
+ * names would let it consume `?code=`/`?error=` params from an unrelated flow
+ * the app runs itself. A namespaced param is proof the redirect came from this
+ * component, which is also what keeps the client's `invalid_flow` detection
+ * meaningful.
+ *
+ * Shared by the component (which writes them) and the browser client (which
+ * reads them) so the two ends can never drift.
+ *
+ * @module
+ */
+
+/** Carries the one-time code the client redeems for a session. */
+export const OAUTH_CODE_PARAM = "convexAuthCode";
+
+/** Carries a normalized {@link OAuthFlowError} code when the flow failed. */
+export const OAUTH_ERROR_PARAM = "convexAuthError";
--- a/packages/core/src/lib/oauthParams.ts
+++ b/packages/core/src/lib/oauthParams.ts
@@ -8,7 +8,7 @@
  * handler is registered unconditionally (every app that uses Convex Auth runs
  * it on every page load), so bare names would let it consume
  * `?code=`/`?error=` params from an unrelated flow the app runs. A
- * `convexAuth`-prefixed param is proof the redirect came from this component,
+ * `convexAuth`-prefixed param signals the redirect came from this component,
  * which is also what keeps the client's `invalid_flow` detection meaningful.
  *
  * Shared by the component (which writes them) and the browser client (which
--- a/packages/core/src/lib/oauthParams.ts
+++ b/packages/core/src/lib/oauthParams.ts
@@ -1,14 +1,15 @@
 /**
- * The query params the OAuth component's callback appends when it redirects the
- * browser back to the app, and that the browser client reads on mount.
+ * URL query param names used by the OAuth component, appended to callback
+ * redirects back to the browser.
  *
- * They are deliberately namespaced rather than the bare `code`/`error` an OAuth
- * callback would normally use. The client's callback handler is registered
- * unconditionally (every app that uses Convex Auth runs it on mount), so bare
- * names would let it consume `?code=`/`?error=` params from an unrelated flow
- * the app runs itself. A namespaced param is proof the redirect came from this
- * component, which is also what keeps the client's `invalid_flow` detection
- * meaningful.
+ * They are deliberately prefixed with `convexAuth` rather than the bare
+ * `code`/`error` an OAuth callback would normally use, so they can't collide
+ * with URL params that an app may use for other reasons. The client's callback
+ * handler is registered unconditionally (every app that uses Convex Auth runs
+ * it on mount), so bare names would let it consume `?code=`/`?error=` params
+ * from an unrelated flow the app runs. A `convexAuth`-prefixed param is proof
+ * the redirect came from this component, which is also what keeps the client's
+ * `invalid_flow` detection meaningful.
  *
  * Shared by the component (which writes them) and the browser client (which
  * reads them) so the two ends can never drift.
--- a/packages/core/src/lib/oauthParams.ts
+++ b/packages/core/src/lib/oauthParams.ts
@@ -6,10 +6,10 @@
  * `code`/`error` an OAuth callback would normally use, so they can't collide
  * with URL params that an app may use for other reasons. The client's callback
  * handler is registered unconditionally (every app that uses Convex Auth runs
- * it on mount), so bare names would let it consume `?code=`/`?error=` params
- * from an unrelated flow the app runs. A `convexAuth`-prefixed param is proof
- * the redirect came from this component, which is also what keeps the client's
- * `invalid_flow` detection meaningful.
+ * it on every page load), so bare names would let it consume
+ * `?code=`/`?error=` params from an unrelated flow the app runs. A
+ * `convexAuth`-prefixed param is proof the redirect came from this component,
+ * which is also what keeps the client's `invalid_flow` detection meaningful.
  *
  * Shared by the component (which writes them) and the browser client (which
  * reads them) so the two ends can never drift.
--- a/packages/core/src/oauth/component/crypto.ts
+++ b/packages/core/src/oauth/component/crypto.ts
@@ -49,16 +49,15 @@ export function decodeJwtPayloadUnverified(
 }
 
 /**
- * Derive an AES-256-GCM key from a raw one-time token. A domain-separated
- * SHA-256 suffices as the KDF: the token is itself 256 bits of CSPRNG
- * output, so no stretching is needed. The `convex-auth-ticket-payload`
- * context string binds derived keys to exactly this use — these helpers are
- * ticket-specific by construction, not general-purpose encryption.
+ * AES-256-GCM key for a ticket payload. The key is only as strong as the
+ * ticket code, which is 256 bits of randomness. The prefix is load-bearing:
+ * the ticket row stores SHA-256(code) as its lookup hash, so without it the
+ * stored hash would be the key.
  */
-async function deriveTicketKey(encodedToken: string): Promise<CryptoKey> {
+async function deriveTicketPayloadKey(ticketCode: string): Promise<CryptoKey> {
   const keyBytes = await crypto.subtle.digest(
     "SHA-256",
-    new TextEncoder().encode(`convex-auth-ticket-payload:${encodedToken}`),
+    new TextEncoder().encode(`convex-auth-ticket-payload:${ticketCode}`),
   );
   return await crypto.subtle.importKey("raw", keyBytes, "AES-GCM", false, [
     "encrypt",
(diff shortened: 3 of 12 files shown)


erquhart:
* **#433** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/433?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#432** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/432?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/432?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#431** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/431?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `reboot`


<h2></h2>

packages/core/src/lib/oauthParams.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * The query params the OAuth component's callback appends when it redirects the
+ * browser back to the app, and that the browser client reads on mount.
+ *
+ * They are deliberately namespaced rather than the bare `code`/`error` an OAuth
```
dowski: nit: "... deliberately prefixed with "convexAuth" ...".

packages/core/src/lib/oauthParams.ts
```diff
@@ -0,0 +1,23 @@
...
+ * callback would normally use. The client's callback handler is registered
+ * unconditionally (every app that uses Convex Auth runs it on mount), so bare
+ * names would let it consume `?code=`/`?error=` params from an unrelated flow
+ * the app runs itself. A namespaced param is proof the redirect came from this
+ * component, which is also what keeps the client's `invalid_flow` detection
+ * meaningful.
```
dowski: Drop the use of "namespaced" below (like most things Claude writes, it sounds fancier than what it is and can be confusing). And maybe state directly that this is to help avoid collisions with other URL params that the app might use.

packages/core/src/lib/oauthParams.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * The query params the OAuth component's callback appends when it redirects the
```
dowski: nit: "The URL query param names ..."

packages/core/src/oauth/component/crypto.ts
```diff
@@ -24,3 +25,87 @@ export async function sha256Base64Url(value: string): Promise<string> {
...
+  ]);
+}
+
+/**
+ * Encrypt a ticket payload with a key derived from the raw one-time token.
+ * Only the token's hash is persisted, and the raw token — the only value the
+ * key can be derived from — travels solely in the callback redirect URL, so
+ * database access alone cannot decrypt the result. Returns the random nonce
+ * followed by the ciphertext, base64url-encoded.
+ */
```
dowski, line 89: Can you describe what a ticket and its payload is here?

dowski: More broadly, I wonder if these functions should be in a reusable place like this, or if they're better situated where they're meant to be used (the HTTP code exchange impl). They can be private as an implementation detail there.

packages/core/src/oauth/component/http.test.ts
```diff
@@ -0,0 +1,565 @@
...
+    const { state } = await startFlow(t);
+    vi.advanceTimersByTime(11 * 60 * 1000);
+    const response = await callback(t, "google", { state, code: "code-1" });
+    expect(redirectParams(response).get(OAUTH_ERROR_PARAM)).toBe("expired");
+  });
+
+  test("a request claimed via the wrong provider's route is refused before any exchange", async () => {
+    const t = setup();
+    const errors = spyConsoleError();
+    const calls = stubFetch({});
```
dowski: What is this doing, as well as the expectation on `calls` below? Is it necessary to stub `fetch` for this test case?

packages/core/src/oauth/component/http.test.ts
```diff
@@ -0,0 +1,565 @@
...
+    const ott = redirectParams(response).get(OAUTH_CODE_PARAM)!;
+    expect(ott).toEqual(expect.any(String));
+    const claimed = await t.mutation(api.provider.claimTicket, {
+      provider: "google",
+      ottHash: await sha256Hex(ott),
+      stateHash,
+    });
+    expect(claimed).not.toBeNull();
+    const payload = JSON.parse(await decryptWithToken(ott, claimed!.payload));
+    expect(payload).toEqual({ claims });
```
dowski: Is this simulating:  
  
1\. a client app receiving the OTT  
2\. the client calling to an exposed function to finish the sign in  
  
?

packages/core/src/oauth/component/http.test.ts
```diff
@@ -0,0 +1,565 @@
...
+    const claimed = await t.mutation(api.provider.claimTicket, {
+      provider: "github",
+      ottHash: await sha256Hex(ott),
+      stateHash,
+    });
+    const payload = JSON.parse(await decryptWithToken(ott, claimed!.payload));
+    expect(payload).toEqual({ userInfoResponses: { user, emails } });
+  });
+
+  test("stale outcome params on redirectTo are replaced, app params kept", async () => {
```
dowski, line 417: Is this testing a case where OAuth is retried from a failed attempt and the redirectTo is built from the current URL with error state in it?

packages/core/src/oauth/component/http.ts
```diff
@@ -0,0 +1,357 @@
...
+  });
+}
+
+/**
+ * 302 back to the app's `redirectTo` with outcome params. Stale outcome params
+ * from a previous attempt are removed first, so a `redirectTo` derived from the
+ * app's current URL can't carry a contradictory outcome. The params are
+ * namespaced ({@link OAUTH_CODE_PARAM}/{@link OAUTH_ERROR_PARAM}) so the app's
+ * always-on client callback handler can tell this redirect apart from an
+ * unrelated `?code=`/`?error=` the app uses for its own purposes.
```
dowski: I don't think it's necessary to repeat this bit about namespacing/prefixing here. Also nothing guarantees that those are the params that are passed in.

packages/core/src/oauth/component/http.ts
```diff
@@ -0,0 +1,357 @@
...
+    };
+  } finally {
+    clearTimeout(timeout);
+  }
+}
+
+/**
+ * Exchange an authorization code for tokens at the provider's token
+ * endpoint. Client credentials are sent in the POST body — empirically the
+ * most compatible style (HTTP Basic has a form-urlencoding ambiguity many
```
dowski: Sort of a general comment in a specific spot.  
  
I suggest removing "Claude-isms" like this. We didn't emperically validate anything, and Claude just loves to overstate/verbosely explain things.  
  
Do a pass over docs and make sure they are useful to human readers and not things that Claude just blabs about to sound smart. :laughing:

packages/core/src/oauth/component/http.ts
```diff
@@ -0,0 +1,357 @@
...
+ *   id_token comes back: `sub` is only unique within an issuer, and a shared
+ *   or multi-tenant token endpoint can serve several.
+ * - `aud` must be exactly CLIENT_ID. Multi-audience tokens are rejected
+ *   because no other audience is trusted.
+ * - `azp`, when present, must be CLIENT_ID.
+ * - `exp` must be in the future.
+ *
+ * Signature verification is deliberately skipped: the token came directly
+ * from the provider's token endpoint over TLS, which OIDC Core sanctions in
+ * place of checking the signature.
```
dowski, line 144: The fetching code should probably assert that the URL is `https`.

packages/core/src/oauth/component/http.ts
```diff
@@ -0,0 +1,357 @@
...
+      { status: 400 },
+    );
+  }
+  // An expired request still knows where the user came from; send them
+  // back to the app instead of stranding them here.
+  if (authRequest.expired) {
+    return redirectToApp(authRequest.redirectTo, {
+      [OAUTH_ERROR_PARAM]: "expired",
+    });
+  }
```
dowski, line 250: Should these errors be logged too?

packages/core/src/oauth/component/schema.ts
```diff
@@ -48,4 +48,39 @@ export default defineSchema({
...
+     * caller-presented state against it, binding completion to the browser
+     * or server that initiated the flow.
+     */
+    stateHash: v.string(),
+    /**
+     * sha256 of the server-minted one-time token. The raw value appears only
+     * in the callback redirect URL and is never stored.
+     */
+    ottHash: v.string(),
+    /** Tickets expire quickly (~2 minutes). */
```
dowski: Maybe just note what this is. The expiration value is controlled elsewhere.

packages/core/src/oauth/component/setup.ts
```diff
@@ -129,6 +131,10 @@ const PROTOCOL_PARAMS = [
  * `<PROVIDER>_CLIENT_ID`/`_SECRET` bindings; register
  * `<site-url><httpPrefix>/<provider>/callback` as the redirect URI with each
  * provider.
+ *
+ * The callback only accepts GET redirects. Providers that POST it
+ * (`response_mode=form_post`, notably Apple when name/email scopes are
```
dowski, line 83: Probably worth a TODO. We're going to want to support Apple at launch.

dowski (changes requested):

packages/core/src/oauth/component/crypto.ts
dowski: I see later they have to be used outside of the HTTP code. So maybe this is okay.

packages/core/src/lib/oauthParams.ts
erquhart: I did actually think namespaced was pretty widely understood, but I'll break it down some more 👍

packages/core/src/lib/oauthParams.ts
erquhart: query param def means url but maybe outside of the web world its not as obvious, will specify url moving forward

packages/core/src/oauth/component/http.ts
erquhart: I actually thought this was helpful - technically http basic is the most spec compliant approach, but in practice using the post body works best with provider implementations. I did read through this stuff! But I can pare things down more if preferred. I used to trim source docs more, but my thinking has shifted toward being more tolerant since 98% of the time the reader will indeed be an llm. Let me know what you think.

packages/core/src/lib/oauthParams.ts
dowski: I don't feel strongly - your call.

packages/core/src/oauth/component/http.ts
dowski: I think I was just confused/wrong here.

packages/core/src/oauth/component/http.ts
dowski: Okay, it's possible I'm just reacting to reading too much Claude output.

packages/core/src/oauth/component/http.test.ts
erquhart: The callback actually uses fetch, so we have to intercept. convex-test doesn't have anything baked in for this (not even sure it should), so we stub it.  This is accepted practice in the docs too.

packages/core/src/oauth/component/http.test.ts
erquhart: Effectively yes - the direct `claimTicket` and decryption at the end is doing those two things. The completion leg is in the next PR, so this test redeems the ticket directly to exercise the ott + state actually decrypting the payload.

packages/core/src/oauth/component/http.test.ts
erquhart: Exactly, makes sure we're dropping stale oauth params from us and preserving everything else.

packages/core/src/oauth/component/crypto.ts
erquhart: Yeah they have to be centralized for reuse

packages/core/src/oauth/component/http.ts
erquhart: I've added console warns, they're somewhat routine so that feels like a good log level

KNOWN_ISSUES.md
```diff
@@ -29,3 +29,15 @@ No action item, just here for awareness.
...
 rejected at setup (`packages/core/src/oauth/component/setup.ts`), so React
 Native apps must return via https universal links / app links.
+
+## Apple sign-in isn't supported yet
+
+Two gaps. The callback route only accepts GET redirects, and Apple POSTs the
+callback (`response_mode=form_post`) whenever name/email scopes are
+requested. And Apple has no static client secret: it requires a short-lived
+ES256 client-secret JWT, signed with a registered key and rotated, where the
+catalogs assume a static `CLIENT_SECRET` binding.
```
dowski, line 39: Could the `CLIENT_SECRET` in the case of Apple be the "registered key" that's used for signing?

packages/core/src/lib/oauthParams.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * URL query param names used by the OAuth component, appended to callback
+ * redirects back to the browser.
+ *
+ * They are deliberately prefixed with `convexAuth` rather than the bare
+ * `code`/`error` an OAuth callback would normally use, so they can't collide
+ * with URL params that an app may use for other reasons. The client's callback
+ * handler is registered unconditionally (every app that uses Convex Auth runs
+ * it on mount), so bare names would let it consume `?code=`/`?error=` params
+ * from an unrelated flow the app runs. A `convexAuth`-prefixed param is proof
```
dowski: nit: let's not say "proof". It's more convention that exists for the reasons your comment already lays out.

packages/core/src/oauth/component/crypto.ts
```diff
@@ -24,3 +25,93 @@ export async function sha256Base64Url(value: string): Promise<string> {
...
+  return Uint8Array.from(binary, (c) => c.charCodeAt(0));
+}
+
+/**
+ * Decode a JWT's payload without verifying its signature. Only safe for
+ * tokens received directly from the issuer over TLS (e.g. an id_token from
+ * the token exchange), where transport authenticates the issuer.
+ */
+export function decodeJwtPayloadUnverified(
+  jwt: string,
```
dowski: Could we require a branded type to prevent this from accidentally being used elsewhere?

packages/core/src/oauth/component/crypto.ts
```diff
@@ -24,3 +25,93 @@ export async function sha256Base64Url(value: string): Promise<string> {
...
+  if (parts.length !== 3) {
+    throw new Error("Malformed JWT");
+  }
+  return JSON.parse(new TextDecoder().decode(base64UrlDecode(parts[1])));
+}
+
+/**
+ * AES-256-GCM key for a ticket payload. The key is only as strong as the
+ * ticket code, which is 256 bits of randomness. The prefix is load-bearing:
+ * the ticket row stores SHA-256(code) as its lookup hash, so without it the
```
dowski: re: the load-bearing prefix: is that referring to the `convex-auth-ticket-payload` in the implementation below? This seems like something that shouldn't be a part of the API doc comment if so.

packages/core/src/oauth/component/http.ts
```diff
@@ -0,0 +1,328 @@
...
+  decodeJwtPayloadUnverified,
+  encryptTicketPayload,
+  generateRandomToken,
+} from "./crypto";
+import { sha256Hex } from "../../lib/crypto";
+import { OAUTH_CODE_PARAM, OAUTH_ERROR_PARAM } from "../../lib/oauthParams";
+
+const http = httpRouter();
+
+const FETCH_TIMEOUT_MS = 30 * 1000;
```
dowski: Any reason why this is so long? Would aborting sooner and letting the user/client manually retry the auth attempt perhaps be better?  
  
I wonder about lots of clients getting stuck waiting for this and potentially blocking other requests (based on backend concurrent function limits or something).

packages/core/src/oauth/component/provider.ts
```diff
@@ -106,3 +111,72 @@ export const claimAuthorizationRequest = internalMutation({
...
+ * caller (the callback) holds the raw ticket code; only its hash is
+ * stored, and the identity payload arrives already encrypted with a key
+ * derived from that code.
+ */
+export const createTicket = internalMutation({
+  args: {
+    providerName: v.string(),
+    stateHash: v.string(),
+    ticketCodeHash: v.string(),
+    payload: v.string(),
```
dowski: Maybe this should be called `encryptedPayload` to make it clear what should be passed in.

packages/core/src/oauth/component/provider.ts
```diff
@@ -106,3 +111,72 @@ export const claimAuthorizationRequest = internalMutation({
...
+export const claimTicket = mutation({
+  args: {
+    providerName: v.string(),
+    ticketCodeHash: v.string(),
+    stateHash: v.string(),
+  },
+  returns: v.union(
+    v.null(),
+    v.object({
+      payload: v.string(),
```
dowski: Same `encryptedPayload` idea.

packages/core/src/oauth/component/schema.ts
```diff
@@ -36,4 +36,42 @@ export default defineSchema({
...
+    /**
+     * The identity the provider attested: JSON of
+     * `{ claims, userInfoResponses }` (id_token claims when the provider
+     * returned one, userinfo responses keyed by the configured endpoint
+     * names; at least one is present), AES-GCM encrypted with a key derived
+     * from the raw ticket code. The raw code is never stored, so
+     * database access alone cannot read the payload, and provider-chosen
+     * JSON keys never become Convex field names.
+     */
+    payload: v.string(),
```
dowski: `encryptedPayload` here too?

dowski (approved):

packages/core/src/oauth/component/http.ts
erquhart: 30 seconds is long but anything meaningfully shorter (like 10s or even 20s) could stop real delays over a terrible connection

erquhart: actually claude corrected me here, this is server to server so 10s is fair (and the sequential block concern is legit)

KNOWN_ISSUES.md
erquhart: yeah that's possible, will update the text here

PR #433, opened 2026-07-28
Add OAuth redemption: completeSignIn, profile mapping contract

erquhart:
- Add `completeSignIn`, the app-side mutation that redeems the callback ticket into a session
- Providers declare a typed `profile` mapping from provider identity to the account profile

### DIFF
--- a/packages/core/src/oauth/component/setup.ts
+++ b/packages/core/src/oauth/component/setup.ts
@@ -271,7 +271,7 @@ export function setupOauth<
       // payload was encrypted under, so decryption only fails on
       // corruption.
       const { claims, userInfoResponses } = JSON.parse(
-        await decryptTicketPayload(args.code, ticket.payload),
+        await decryptTicketPayload(args.code, ticket.encryptedPayload),
       ) as {
         claims: OidcClaims | undefined;
         userInfoResponses: UserInfo | undefined;
--- a/packages/core/src/oauth/component/setup.ts
+++ b/packages/core/src/oauth/component/setup.ts
@@ -14,8 +14,8 @@ import {
 import { sha256Hex } from "../../lib/crypto";
 
 /**
- * Standard OIDC id_token claims, loosely typed: the well-known ones are
- * named, everything else comes through the index signature.
+ * Standard OIDC id_token claims. The well-known ones are typed; any other
+ * claim the provider includes is present but untyped (`unknown`).
  */
 export type OidcClaims = {
   sub: string;
@@ -32,18 +32,33 @@ export type OidcClaims = {
  * configured (`undefined` unless the catalog sets `userInfoEndpoints`) — to the
  * account identity used at redemption. `id` becomes the provider account id.
  * Supplied by each provider's catalog.
+ *
+ * `UserInfo` is the catalog's declared shape for the userinfo responses,
+ * keyed like `userInfoEndpoints`. It types what the provider is trusted to
+ * return — the responses are provider-attested JSON and are not validated
+ * against it at runtime.
  */
-export type OauthProfile = (
-  claims: OidcClaims | undefined,
-  // `any` so mappings can dig into responses without casting.
+export type OauthProfile<
+  // `any` default so unparameterized mappings can dig into responses
+  // without casting.
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  userInfoResponses: Record<string, any> | undefined,
+  UserInfo extends Record<string, unknown> = Record<string, any>,
+> = (
+  claims: OidcClaims | undefined,
+  userInfoResponses: UserInfo | undefined,
 ) => { id: string; [key: string]: unknown };
 
 /**
  * Provider-defined config for interacting with an individual OAuth provider.
+ *
+ * `UserInfo` declares the shape of the userinfo responses, tying
+ * `userInfoEndpoints` keys to what `profile` receives (see
+ * {@link OauthProfile}).
  */
-export type OauthCatalog = {
+export type OauthCatalog<
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  UserInfo extends Record<string, unknown> = Record<string, any>,
+> = {
   /**
    * The provider's authorization endpoint (a full URL), e.g. Google's
    * `https://accounts.google.com/o/oauth2/v2/auth`.
@@ -65,9 +80,9 @@ export type OauthCatalog = {
    * Endpoints (full URLs) the callback fetches with the access token (GET
    * with a bearer token), keyed by the name the `profile` mapping reads
    * each response under. Present for providers whose identity comes from
-   * userinfo (GitHub).
+   * userinfo (GitHub). The keys must match `UserInfo`'s.
    */
-  userInfoEndpoints?: Record<string, string>;
+  userInfoEndpoints?: { [K in keyof UserInfo & string]: string };
   /** Scopes to request. */
   scopes: string[];
   /**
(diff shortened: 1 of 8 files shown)


erquhart:
* **#433** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/433?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/433?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#458** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/458?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `reboot`


<h2></h2>

packages/core/src/oauth/component/setup.ts
```diff
@@ -1,17 +1,52 @@
...
+  decryptWithToken,
+  generateRandomToken,
+  sha256Base64Url,
+} from "./crypto";
 import { sha256Hex } from "../../lib/crypto";
 import { capitalize } from "../../lib/strings";
 
+/**
+ * Standard OIDC id_token claims, loosely typed: the well-known ones are
+ * named, everything else comes through the index signature.
```
dowski: What is "the index signature"?

packages/core/src/oauth/component/setup.ts
```diff
@@ -1,17 +1,52 @@
...
+ * (`undefined` for non-OIDC providers) and userinfo responses keyed as
+ * configured (`undefined` unless the catalog sets `userInfoEndpoints`) — to the
+ * account identity used at redemption. `id` becomes the provider account id.
+ * Supplied by each provider's catalog.
+ */
+export type OauthProfile = (
+  claims: OidcClaims | undefined,
+  // `any` so mappings can dig into responses without casting.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  userInfoResponses: Record<string, any> | undefined,
```
dowski: I wonder if this could be constrained to actual response shape based on the provider.

packages/core/src/oauth/component/setup.ts
```diff
@@ -122,10 +159,13 @@ const PROTOCOL_PARAMS = [
  * 2. The provider redirects back to the component's HTTP callback
  *    (`<site><httpPrefix>/<provider>/callback`), which claims the request,
  *    exchanges the code, and mints a one-time ticket.
+ * 3. `completeSignIn` (here): the client presents the one-time code from the
```
dowski, line 131: Another Claude-ism: it loves to note things about the temporal/in-progress work in doc comments that are meant to stand alone. Drop "(here)" at least.

dowski (changes requested):
Can you include tests for `completeSignIn`?

packages/core/src/oauth/component/setup.ts
erquhart: a really bad way of describing the loose string: unknown type, will improve this

packages/core/src/oauth/component/setup.ts
erquhart: I take the "here" to literally be "this happens here, on our side" for steps 1 and 3 as opposed to step 2, which happens on the provider side. Super non-specific, though, can improve this.

dowski: Oh you're right! I see "(here)" in step 1 too. My bad, my bad.

packages/core/src/oauth/component/setup.ts
erquhart: really good call, adding this

packages/core/src/oauth/component/setup.ts
```diff
@@ -1,14 +1,65 @@
...
+ * `UserInfo` is the catalog's declared shape for the userinfo responses,
+ * keyed like `userInfoEndpoints`. It types what the provider is trusted to
+ * return. The responses are provider-attested JSON and are not validated
+ * against it at runtime.
+ */
+export type OauthProfile<
+  // `any` default so unparameterized mappings can dig into responses
+  // without casting.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  UserInfo extends Record<string, unknown> = Record<string, any>,
```
dowski: Do you think we could type any of this more tightly? When would there be an unparameterized mapping?

packages/core/src/oauth/component/setup.ts
```diff
@@ -1,14 +1,65 @@
...
  * Provider-defined config for interacting with an individual OAuth provider.
+ *
+ * `UserInfo` declares the shape of the userinfo responses, tying
+ * `userInfoEndpoints` keys to what `profile` receives (see
+ * {@link OauthProfile}).
  */
-export type OauthCatalog = {
+export type OauthCatalog<
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  UserInfo extends Record<string, unknown> = Record<string, any>,
```
dowski: Similar thought re: typing here.

packages/core/src/oauth/component/setup.ts
```diff
@@ -78,15 +131,21 @@ function parseUrl(value: string): URL | null {
...
  *
  * The callback only accepts GET redirects. Providers that POST it
  * (`response_mode=form_post`, notably Apple when name/email scopes are
  * requested) are not supported yet.
  * TODO: support response_mode=form_post (Apple) before launch.
  */
-export function setupOauth(
+export function setupOauth<
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  UserInfo extends Record<string, unknown> = Record<string, any>,
```
dowski: Also here, can we get rid of `any`?

packages/core/src/oauth/component/setup.ts
erquhart: Yeah fair, it was originally built for open usage against arbitrary providers but we're narrowing in on specific providers, so may as well go all the way. Going to probably just type this stuff out of the catalogs in pr 4 directly and get rid of OauthProfile.

erquhart: actually going to just minimally do what you're asking here, I have proposals for doing it much better but we need the base flow landed so I'll propose more after that

dowski (approved):

erquhart:
@dowski couple new commits added, they're small

dowski (approved):

packages/core/src/oauth/component/setup.ts
```diff
@@ -21,25 +68,28 @@ export type OauthCatalog = {
   tokenEndpoint: string;
   /**
    * Expected `iss` (the OIDC issuer claim) of the provider's id_tokens, e.g.
-   * Google's `https://accounts.google.com`. Present for OIDC providers;
-   * absent for plain-OAuth providers (e.g., GitHub), where identity comes from
-   * userinfo.
+   * Google's `https://accounts.google.com`. Some providers document more than
+   * one form (Google also uses `accounts.google.com`); an array accepts any
+   * of them. Present for OIDC providers; absent for plain-OAuth providers
+   * (e.g., GitHub), where identity comes from userinfo.
```
dowski, line 74: I guess one thing I wonder is how this aligns with the OIDC spec. This would allow two complete different issuer hosts. Is that allowed/expected or should we just loosen matching to not require the protocol but still match on the host?

dowski (approved):

packages/core/src/oauth/component/setup.ts
erquhart: This doesn't violate spec, we just need to catalog which issuers are trusted for a given provider, and some have more than one. Loosening to host only would actually be more concerning as the spec doesn't really have a near-match concept at all. jose's jwtVerify() also accepts string or string[] for issuer, so this isn't novel on our part.

dowski: Cool, thanks for the info. Sounds like this is a solid approach.

dowski (approved):

PR #439, opened 2026-08-09
return null on pre-verification code errors

erquhart:
- return null on pre-verification code errors
- avoid consuming verification codes on early failures that shouldn't lead to consumption


src/server/implementation/mutations/verifyCodeAndSignIn.ts
dowski, line 144: I wonder if we should delete the code in this case - the code is no good at this point.

dowski (approved):

PR #445, opened 2026-08-11
Erase expired passkey challenges in the background

Nicolapps:
The current implementation of the passkey component has a `challenges` table. Challenges are only valid for 10 minutes, and the current implementation never deletes challenges that were started but then never claimed.

This PR fixes the row leak problem by deleting expired challenges through [`@convex-dev/batch-worker`](https://www.npmjs.com/package/@convex-dev/batch-worker).

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -57,7 +57,7 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
-    "@convex-dev/batch-worker": "^0.2.0",
+    "@convex-dev/batch-worker": "^0.3.0",
     "@convex-dev/rate-limiter": "^0.3.2",
     "argon2id-wasm": "0.0.0-alpha.1",
     "commander": "^14.0.2",
@@ -70,8 +70,8 @@
     "@testing-library/react": "^16.3.2",
     "@types/node": "^24.12.4",
     "@types/react": "^19.2.17",
-    "convex": "^1.41.0",
-    "convex-test": "^0.0.53",
+    "convex": "^1.43.0",
+    "convex-test": "^0.0.55",
     "hash-wasm": "^4.12.0",
     "jsdom": "^29.1.1",
     "next": "^16.0.0",
@@ -82,7 +82,7 @@
     "vitest": "^4.1.0"
   },
   "peerDependencies": {
-    "convex": "^1.41.0",
+    "convex": "^1.43.0",
     "next": "^16.0.0",
     "react": "^18.0.0 || ^19.0.0"
   },
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -57,6 +57,7 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
+    "@convex-dev/batch-worker": "^0.2.0",
     "@convex-dev/rate-limiter": "^0.3.2",
     "argon2id-wasm": "0.0.0-alpha.1",
     "commander": "^14.0.2",
--- a/packages/core/src/components/passkey/cleanup.ts
+++ b/packages/core/src/components/passkey/cleanup.ts
@@ -47,10 +47,12 @@ export const getExpiredChallenges = internalQuery({
     // the batch erases.
     const expired = await ctx.db
       .query("challenges")
-      .withIndex("by_creation_time", (q) =>
-        // The boundary matches `isChallengeExpired`: at exactly the TTL, a
-        // challenge is expired. A wake-up at the deadline thus finds work.
-        q.lte("_creationTime", now - CHALLENGE_TTL_MS),
+      .withIndex(
+        "by_creation_time",
+        (q) =>
+          // The boundary matches `isChallengeExpired`: at exactly the TTL, a
+          // challenge is expired. A wake-up at the deadline thus finds work.
+          q.lte("_creationTime", now - CHALLENGE_TTL_MS), // FIXME: we could consider using the batch worker cursor here to avoid tombstones
       )
       .take(BATCH_SIZE);
     if (expired.length > 0) {
--- a/packages/core/src/components/passkey/cleanup.ts
+++ b/packages/core/src/components/passkey/cleanup.ts
@@ -1,0 +1,90 @@
+import { v } from "convex/values";
+import { ping, vBatchQueryArgs, vBatchResult } from "@convex-dev/batch-worker";
+import { components, internal } from "./_generated/api";
+import { internalMutation, internalQuery } from "./_generated/server";
+import type { MutationCtx } from "./_generated/server";
+import { CHALLENGE_TTL_MS } from "./validation";
+import { isChallengeExpired } from "./helpers";
+
+// The name of the batch worker loop that erases expired challenges. Only one
+// loop is necessary, so the name is constant.
+export const WORKER_NAME = "expiredChallenges";
+
+// How many challenges one worker mutation erases. The worker runs again
+// immediately while more expired rows stay, so this value only sets the
+// size of each transaction.
+const BATCH_SIZE = 64;
+
+/**
+ * Make sure that the background cleanup loop runs.
+ *
+ * Call this function after you write a challenge. The call is cheap: it does
+ * nothing while the loop already runs. The loop stops when no expired
+ * challenge stays.
+ */
+export async function scheduleChallengeCleanup(ctx: MutationCtx) {
+  await ping(ctx, components.batchWorker, {
+    name: WORKER_NAME,
+    workQuery: internal.cleanup.getExpiredChallenges,
+    workerMutation: internal.cleanup.deleteExpiredChallenges,
+  });
+}
+
+/**
+ * Find the next batch of expired challenges.
+ *
+ * When no challenge is expired, the loop goes idle. If unexpired challenges
+ * stay, the loop also gets a timeout: it then wakes up again when the oldest
+ * of them expires, even if no new ceremony starts.
+ */
+export const getExpiredChallenges = internalQuery({
+  args: vBatchQueryArgs,
+  returns: vBatchResult(v.object({ ids: v.array(v.id("challenges")) })),
+  handler: async (ctx) => {
+    const now = Date.now();
+    // One indexed read answers both questions: which rows are expired, and
+    // when the oldest row expires. The index sorts by `createdAt`, so the
+    // batch holds the oldest rows.
+    const oldestRows = await ctx.db
+      .query("challenges")
+      .withIndex("by_createdAt")
+      .take(BATCH_SIZE);
+    const expired = oldestRows.filter((row) => isChallengeExpired(row, now));
+    if (expired.length > 0) {
+      return {
+        kind: "work" as const,
+        batch: { ids: expired.map((row) => row._id) },
+      };
+    }
+    // Nothing to erase now. The oldest remaining challenge is the first one
+    // that expires.
+    const oldest = oldestRows[0];
+    if (oldest === undefined) {
+      return { kind: "idle" as const };
+    }
+    return {
+      kind: "idle" as const,
+      timeoutMs: Math.max(0, oldest.createdAt + CHALLENGE_TTL_MS - now),
+    };
+  },
+});
+
+/**
+ * Erase one batch of expired challenges.
+ *
+ * The worker owns the cleanup of the rows that it processes. A challenge can
+ * also be consumed between the query and this mutation, thus the rows that no
+ * longer exist are ignored.
+ */
+export const deleteExpiredChallenges = internalMutation({
+  args: { ids: v.array(v.id("challenges")) },
+  returns: v.null(),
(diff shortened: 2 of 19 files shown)


Nicolapps:
* **#445** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/445?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 3  dependent PRs ([#460](https://github.com/get-convex/convex-auth/pull/460) <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/460?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#489](https://github.com/get-convex/convex-auth/pull/489) <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/489?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#490](https://github.com/get-convex/convex-auth/pull/490) <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/490?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/445?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `reboot`


<h2></h2>

packages/core/src/components/passkey/schema.ts
ianmacartney: how does this differ from _creationTime?

packages/core/src/components/passkey/cleanup.ts
```diff
@@ -0,0 +1,90 @@
...
+  returns: vBatchResult(v.object({ ids: v.array(v.id("challenges")) })),
+  handler: async (ctx) => {
+    const now = Date.now();
+    // One indexed read answers both questions: which rows are expired, and
+    // when the oldest row expires. The index sorts by `createdAt`, so the
+    // batch holds the oldest rows.
+    const oldestRows = await ctx.db
+      .query("challenges")
+      .withIndex("by_createdAt")
+      .take(BATCH_SIZE);
```
ianmacartney: If you use 0.3.0 you can have it pass you a cursor to use and return, so you can skip tombstones from all the previously deleted challenges. maybe overkill given how few challenges there will likely be.

packages/core/src/components/passkey/cleanup.ts
```diff
@@ -0,0 +1,90 @@
...
+    const expired = oldestRows.filter((row) => isChallengeExpired(row, now));
```
ianmacartney: why not encode `.lt("createdAt", now - CHALLENGE_TTL_MS)` into the query?

packages/core/src/components/passkey/cleanup.ts
```diff
@@ -0,0 +1,90 @@
...
+ * The worker owns the cleanup of the rows that it processes. A challenge can
+ * also be consumed between the query and this mutation, thus the rows that no
+ * longer exist are ignored.
+ */
+export const deleteExpiredChallenges = internalMutation({
+  args: { ids: v.array(v.id("challenges")) },
+  returns: v.null(),
+  handler: async (ctx, { ids }) => {
+    for (const id of ids) {
+      if ((await ctx.db.get("challenges", id)) !== null) {
```
ianmacartney: the get here is redundant / wasteful - the delete will already take a read dependency, and this runs on the same db snapshot as the query, so anything you return from the query will exist here

packages/core/src/components/passkey/cleanup.ts
```diff
@@ -0,0 +1,90 @@
...
+import { isChallengeExpired } from "./helpers";
+
+// The name of the batch worker loop that erases expired challenges. Only one
+// loop is necessary, so the name is constant.
+export const WORKER_NAME = "expiredChallenges";
+
+// How many challenges one worker mutation erases. The worker runs again
+// immediately while more expired rows stay, so this value only sets the
+// size of each transaction.
+const BATCH_SIZE = 64;
```
ianmacartney: nit: this could safely be 1024 I think - each document is tiny, and if the query bounds exclude un-expired entries, it will generally be small.

packages/core/package.json
```diff
@@ -56,6 +56,7 @@
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
+    "@convex-dev/batch-worker": "^0.2.0",
```
ianmacartney: update to 0.3.0

ianmacartney (commented):
Love to see batch worker getting used!

packages/core/src/components/passkey/cleanup.ts
Nicolapps: Thanks! I saw my agent had written this and I didn't question it further. I had assumed that the mutation might be running in a transaction that's separate from the query, and that the agent was trying to avoid it. Will fix

packages/core/package.json
Nicolapps: Done ✅

packages/core/src/components/passkey/schema.ts
Nicolapps: Sounds good, I removed them. Often I'm skeptical about using _creationTime because it can make it harder to import data with custom dates, but in this case it's fine to force using the real creation time

packages/core/src/components/passkey/cleanup.ts
Nicolapps: While working on upstack PRs, I told my coding agent at some point to create a `isChallengeExpired` function and use it “everywhere,” and it took it too literally.

I saw that the agent had made changes to this PR (this is why I put it back to drafts) but I had not reviewed the changes yet. Next time I’ll make things clearer and temporarily remove the review request too

packages/core/src/components/passkey/cleanup.ts
```diff
@@ -0,0 +1,94 @@
...
+    // that expires, thus the loop wakes up again at its deadline even if no
+    // new ceremony starts.
+    const oldest = await ctx.db.query("challenges").order("asc").first();
+    if (oldest === null) {
+      return { kind: "idle" as const };
+    }
+    return {
+      kind: "idle" as const,
+      timeoutMs: Math.max(0, oldest._creationTime + CHALLENGE_TTL_MS - now),
+    };
```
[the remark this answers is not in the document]
Nicolapps: image

packages/core/src/components/passkey/schema.ts
```diff
@@ -40,9 +42,7 @@ export default defineSchema({
         // The field is not set for discoverable-credential ceremonies. In
         // that flow, the assertion identifies the user.
         userId: v.optional(v.string()),
-        createdAt: v.number(),
       }),
     ),
-    // TODO(nicolas) Automatically erase expired challenges in the background
```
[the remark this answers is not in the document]
Nicolapps: This is very fine

packages/core/src/components/passkey/cleanup.ts
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Find the next batch of expired challenges.
+ *
+ * When no challenge is expired, the loop goes idle. If unexpired challenges
+ * stay, the loop also gets a timeout: it then wakes up again when the oldest
+ * of them expires, even if no new ceremony starts.
+ */
+export const getExpiredChallenges = internalQuery({
+  args: vBatchQueryArgs,
+  returns: vBatchResult(v.object({ ids: v.array(v.id("challenges")) })),
```
ianmacartney, line 42: nit: These helpers are deprecated. You can now create the args & return validators for both the query and mutation via one call to `defineBatchWorkerValidators`. That will also allow you to specify a validator for your cursor type (v.number() in your case).

packages/core/src/components/passkey/cleanup.ts
```diff
@@ -0,0 +1,94 @@
...
+    // the rows by their expiry too. The bound keeps the read to the rows that
+    // the batch erases.
+    const expired = await ctx.db
+      .query("challenges")
+      .withIndex(
+        "by_creation_time",
+        (q) =>
+          // The boundary matches `isChallengeExpired`: at exactly the TTL, a
+          // challenge is expired. A wake-up at the deadline thus finds work.
+          q.lte("_creationTime", now - CHALLENGE_TTL_MS), // FIXME: we could consider using the batch worker cursor here to avoid tombstones
```
ianmacartney, line 55:
```suggestion
          q.gte("_creationTime", args.cursor ?? 0).lt("_creationTime", now - CHALLENGE_TTL_MS),
```

packages/core/src/components/passkey/cleanup.ts
```diff
@@ -0,0 +1,94 @@
...
+ * Find the next batch of expired challenges.
+ *
+ * When no challenge is expired, the loop goes idle. If unexpired challenges
+ * stay, the loop also gets a timeout: it then wakes up again when the oldest
+ * of them expires, even if no new ceremony starts.
+ */
+export const getExpiredChallenges = internalQuery({
+  args: vBatchQueryArgs,
+  returns: vBatchResult(v.object({ ids: v.array(v.id("challenges")) })),
+  handler: async (ctx) => {
```
ianmacartney, line 43:
```suggestion
  handler: async (ctx, args) => {
```

packages/core/src/components/passkey/cleanup.ts
```diff
@@ -0,0 +1,94 @@
...
+        "by_creation_time",
+        (q) =>
+          // The boundary matches `isChallengeExpired`: at exactly the TTL, a
+          // challenge is expired. A wake-up at the deadline thus finds work.
+          q.lte("_creationTime", now - CHALLENGE_TTL_MS), // FIXME: we could consider using the batch worker cursor here to avoid tombstones
+      )
+      .take(BATCH_SIZE);
+    if (expired.length > 0) {
+      return {
+        kind: "work" as const,
```
ianmacartney, line 60:
```suggestion
        kind: "work" as const,
        cursor: expired.at(-1)!._creationTime
```

ianmacartney (approved):

packages/core/src/components/passkey/cleanup.ts
Nicolapps: Thanks for letting me know.

Should `vBatchQueryArgs` and `vBatchResult` be marked as `@deprecated` in the TSDoc?

It also looks like the plugin README (https://www.convex.dev/components/batch-worker) mentions both

packages/core/src/components/passkey/cleanup.ts
Nicolapps: Done in https://github.com/get-convex/convex-auth/pull/487

packages/core/src/components/passkey/cleanup.ts
Nicolapps: Done in https://github.com/get-convex/convex-auth/pull/487

packages/core/src/components/passkey/cleanup.ts
Nicolapps: Done in https://github.com/get-convex/convex-auth/pull/487 (with a small modification)

packages/core/src/components/passkey/cleanup.ts
Nicolapps: Done in https://github.com/get-convex/convex-auth/pull/487

packages/core/src/components/passkey/cleanup.ts
ianmacartney: Hmm, they are @deprecated in the TSDocs already, and it's not in the README, but I do see that the AI-generated summary on the page you linked mentions them. I'll update that.
I wonder why it didn't show up as deprecated for you... I'm not going to dig deeper but if you think of anything to improve here let me know!
Screenshot 2026-08-18 at 2 47 10 PM

Nicolapps: Thanks, I can confirm they also appear as deprecated in my editor now. I wonder if Zed could have been caching the TSDoc of an older version

ianmacartney:
thanks for confirming!

packages/core/src/components/passkey/cleanup.ts
```diff
@@ -0,0 +1,94 @@
...
+ * Call this function after you write a challenge. The call is cheap: it does
+ * nothing while the loop already runs. The loop stops when no expired
+ * challenge stays.
+ */
+export async function scheduleChallengeCleanup(ctx: MutationCtx) {
+  await ping(ctx, components.batchWorker, {
+    name: WORKER_NAME,
+    workQuery: internal.cleanup.getExpiredChallenges,
+    workerMutation: internal.cleanup.deleteExpiredChallenges,
+  });
```
ianmacartney, line 30: Once https://github.com/get-convex/batch-worker/pull/16 is shipped, you could set a debounceMs here to be `CHALLENGE_TTL_MS` , which is a minor optimization, but skips running the worker until there's actually a chance that something is worth doing. For now, doing so might perpetually kick out the start time when it's scheduled.

Nicolapps: Thanks!

PR #440, opened 2026-08-10
Unit tests for the passkey component

Nicolapps:

PR #438, opened 2026-08-06
Vendor Oslo dependencies

Nicolapps:
Oslo [is deprecated](https://pilcrowonpaper.com/blog/18), so I’m going to vendor Pilcrow’s crypto/Webauthn implementations into the package itself.

PR #437, opened 2026-08-06
Passkey component

Nicolapps:
This PR adds a new passkey component. This component is fully neutral to the authentication flow being used: it only includes methods to:
- register a new passkey
- authenticate using existing passkeys
- list existing passkeys for a user and delete one

PR #446, opened 2026-08-11
Add a username component

Nicolapps:
Right now, the “username + password” flow stores the username as an account ID in the core `accounts` table.

This is not conceptually right: for instance, in an app that uses both passwords and passkeys, the username shouldn’t “belong” to one of the two providers. Moreover, the current structure doesn't work well for apps that want to let users log in with their username _or_ their email address.

This PR is adding a new username component that is very simple: it only stores the user ID ↔ username mapping, and normalizes the value for allowing case-insensitive lookups.

The `accounts` table no longer needs to store an account ID, so I’m using `""` as a value for the password provider.

Validation logic:
- The username component doesn't implement opinionated validation rules. It implements a few validation rules that all apps will want (e.g. `1 ≤ length ≤ 100`, no control characters…), but doesn't do more opinionated things (like only accepting ASCII characters). We can add additional constraints in provider code (with possibly customization options).

Plans for future extensibility:
- We will other components that can each implement their own user lookup mechanism (e.g. email, phone number)
- If users want to implement flows such as “log in through username _or_ email address _or_ phone number”, they can do so in app code (with a sign in function that queries the relevant components, or multiple of them until it finds a match)

### DIFF
--- a/packages/core/src/components/password/setup.ts
+++ b/packages/core/src/components/password/setup.ts
@@ -4,7 +4,6 @@ import { defineProvider, vTokenBundle } from "../../lib/types";
 import type { ComponentApi } from "./_generated/component.js";
 import type { ComponentApi as UsernameComponentApi } from "../username/_generated/component.js";
 import {
-  normalizeUsername,
   setUsernameUserError,
   validateUsernameFormat,
 } from "../username/validation";
@@ -33,6 +32,10 @@ export type UsernamePasswordOptions = {
 
 // TODO: derive this from the component mount path rather than hardcoding it.
 const PROVIDER_NAME = "password";
+
+// A given user has only zero or one “provider account ID”,
+// so there is no need for having a value here.
+const PROVIDER_ACCOUNT_ID = "";
 
 const signInResult = v.union(
   v.object({ success: v.literal(true), tokens: vTokenBundle }),
@@ -137,7 +140,7 @@ export const UsernamePassword = defineProvider({
           // username. Give the account a stable, opaque id instead.
           const tokens = await completeSignIn(ctx, {
             provider: PROVIDER_NAME,
-            providerAccountId: normalizeUsername(username),
+            providerAccountId: PROVIDER_ACCOUNT_ID,
             profile: { username },
           });
 
@@ -214,7 +217,7 @@ export const UsernamePassword = defineProvider({
           // no longer needs to be keyed by the username.
           const tokens = await completeSignIn(ctx, {
             provider: PROVIDER_NAME,
-            providerAccountId: normalizeUsername(username),
+            providerAccountId: PROVIDER_ACCOUNT_ID,
             profile: { username },
           });
           return { success: true, tokens };
--- a/packages/core/src/components/password/setup.ts
+++ b/packages/core/src/components/password/setup.ts
@@ -35,7 +35,7 @@ const PROVIDER_NAME = "password";
 
 // A given user has only zero or one “provider account ID”,
 // so there is no need for having a value here.
-const PROVIDER_ACCOUNT_ID = "";
+const EMPTY_PROVIDER_ACCOUNT_ID = "";
 
 const signInResult = v.union(
   v.object({ success: v.literal(true), tokens: vTokenBundle }),
@@ -140,7 +140,7 @@ export const UsernamePassword = defineProvider({
           // username. Give the account a stable, opaque id instead.
           const tokens = await completeSignIn(ctx, {
             provider: PROVIDER_NAME,
-            providerAccountId: PROVIDER_ACCOUNT_ID,
+            providerAccountId: EMPTY_PROVIDER_ACCOUNT_ID,
             profile: { username },
           });
 
@@ -217,7 +217,7 @@ export const UsernamePassword = defineProvider({
           // no longer needs to be keyed by the username.
           const tokens = await completeSignIn(ctx, {
             provider: PROVIDER_NAME,
-            providerAccountId: PROVIDER_ACCOUNT_ID,
+            providerAccountId: EMPTY_PROVIDER_ACCOUNT_ID,
             profile: { username },
           });
           return { success: true, tokens };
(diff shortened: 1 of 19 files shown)


Nicolapps:
* **#446** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/446?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#448](https://github.com/get-convex/convex-auth/pull/448) <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/448?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#450](https://github.com/get-convex/convex-auth/pull/450) <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/450?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/446?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#440** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/440?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#438** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/438?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#437** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/437?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `reboot`


<h2></h2>

packages/core/src/components/password/setup.ts
```diff
@@ -17,15 +22,20 @@ export type UsernamePasswordOptions = {
    * recipe drives its `setPassword` / `verifyPassword` mutations.
    */
   component: ComponentApi;
+  /**
+   * The mounted username component (`components.authUsername`). The recipe
```
dowski, line 26: Just `components.username`, right?

packages/core/src/components/password/setup.ts
```diff
@@ -17,15 +22,20 @@ export type UsernamePasswordOptions = {
...
 // TODO: derive this from the component mount path rather than hardcoding it.
 const PROVIDER_NAME = "password";
 
-// Used to compare usernames case-insensitively.
-function normalizeUsername(username: string): string {
-  return username.toLowerCase();
-}
+// A given user has only zero or one “provider account ID”,
+// so there is no need for having a value here.
+const PROVIDER_ACCOUNT_ID = "";
```
dowski: Maybe `EMPTY_PROVIDER_ACCOUNT_ID`?

packages/core/src/components/username/validation.ts
```diff
@@ -0,0 +1,170 @@
...
+
+// The component applies only the rules that are correct for all applications.
+// There is no maximum length, and characters of all scripts are permitted:
+// the component can be used in an international application.
+//
+// The component rejects only usernames that are clearly malformed.
+//
+// An application that wants more strict rules (only letters and digits, a
+// minimum length of 3, a list of reserved names, …) applies them before it
+// calls `setUsername`.
```
dowski, line 11: Practically, I think by default we should constrain usernames further than what's allowed here.

Something like:

https://knowledge.workspace.google.com/admin/users/name-guidelines-for-users-and-groups

should be the default for the `UsernamePassword` provider.

We could allow apps to pass in a custom validation function that overrides the `UsernamePassword` provider defaults.  
  
BetterAuth allows customization like this:  
  
https://better-auth.com/docs/plugins/username#options

packages/core/src/components/username/validation.ts
```diff
@@ -0,0 +1,170 @@
...
+    minimumLength: v.number(),
+  }),
+  // The username starts or ends with a space or a different whitespace
+  // character.
+  v.object({ error: v.literal("USERNAME_HAS_SURROUNDING_WHITESPACE") }),
+  // The username contains a character that is not permitted. See
+  // `hasDisallowedCharacter` for the list.
+  v.object({ error: v.literal("USERNAME_HAS_INVALID_CHARACTERS") }),
+  // A different user already has this username.
+  v.object({ error: v.literal("USERNAME_TAKEN") }),
```
dowski, line 31: Could we have a value for failing app validation too?

packages/core/src/components/username/validation.ts
```diff
@@ -0,0 +1,170 @@
+import { Infer, v } from "convex/values";
+
+// The component applies only the rules that are correct for all applications.
+// There is no maximum length, and characters of all scripts are permitted:
+// the component can be used in an international application.
+//
+// The component rejects only usernames that are clearly malformed.
```
dowski, line 7: These seem like really good backstop rules!

packages/core/src/components/username/validation.ts
dowski: Note that these constraints should be on sign up, not sign in.

packages/core/src/components/password/setup.ts
Nicolapps: @cursor Please implement this change

packages/core/src/components/password/setup.ts
Nicolapps: I think we want to prefix all auth components names with `auth`, because it will make the DX easier (e.g. auth components would be grouped in the dashboard)

packages/core/src/components/username/validation.ts
Nicolapps: Yes, I plan to do something like this for the provider itself

packages/core/src/components/username/validation.ts
Nicolapps: I think this additional error should be added in the app code. It could allow custom providers to have more precise error codes for the errors they are adding, and also remove the need for the frontend to remove error codes that are not used in practice

packages/core/src/components/password/setup.ts
dowski: Yep sounds good - I wrote this just based on the path name being `username` but I see it's `authUsername` in the config now.

dowski (approved):

Nicolapps:
### Merge activity

* **Aug 12, 10:10 PM UTC**: @Nicolapps merged this pull request with [Graphite](https://app.graphite.com/github/pr/get-convex/convex-auth/446).

PR #460, opened 2026-08-13
Passkey component: add an indirection for the ID used in the passkey (“handle”)

Nicolapps:
This PR adds a level of indirection for the IDs used in passkeys. Instead of associating the passkeys to the user ID (`Id<"users">`), we now generate a random 64-bytes value (the “handle”) that is associated with a user.

## Why are we doing this?
When creating a passkey in an authenticator, we need to provide a [user handle](https://www.w3.org/TR/webauthn-3/#user-handle) ([`PublicKeyCredentialUserEntity.id`](https://www.w3.org/TR/webauthn-3/#dom-publickeycredentialcreationoptions-user)), which identifies the account this passkey is authenticated with. This value must be unique for every user, and not contain any personally identifying information such as a email address or username. Moreover, the WebAuthn spec indicates that the user handle “ought to be the same for all credentials registered to the same user account.”

Initially, my intuition was to use the user ID (`Id<"users">`) as a user handle. However, this doesn’t work well for auth flows where the passkey is registered at the same time as the user creation (e.g. username + password). If we were using the user ID as the handle, it would mean that we would need to create the user before registering the first passkey. This situation is worse because it creates a temporary user where a user has no identity[^1] or auth methods[^2], and this user is never cleaned up if the registration process does not help.

[^1]: We don’t want to register the identity before the passkey creation challenge is over, otherwise we could have a sitaution where a username is registered and no one is able to access it (e.g. because the user was not able to finish the passkey registration).
[^2]: Because the passkey has not been registered at this point.

## How are handles implemented in practice?
The passkey component has a new `handles` table that links handles to users. Handles are 64-byte random values, which is what the WebAuthn spec [recommends doing](https://www.w3.org/TR/webauthn-3/#sctn-user-handle-privacy).

There are now two supported flows to register a passkey:
* The passkey is added to an existing user (`startRegistration` called with a `userId` parameter).
  * If it’s the first time this user receives a passkey, we generate a random handle.
  * If the user already had passkeys in the past, we reuse their previous handle.
* The passkey registration is for a user that isn’t created yet (`startRegistration` with no `userId`)
  * We create a handle that’s not associated with a user ID and store it in `handles`
  * When the passkey registration completes, we store the new user ID in the `handles` row. In this flow, the user is expected to be new at this point, so if the user already has a handle we throw an error.

When a passkey is deleted from the component, we don’t delete the handle (even if the user has no passkeys anymore), because an authenticator might still store an expired passkey that contains the handle. However, we still want to clean up handles for users that are deleted: this is done through the new `deleteUser` mutation, which deletes all passkeys for a user and also their handle.

## Are handles just another kind of user ID?
In design discussions for Convex Auth v2, a design goal I had is to avoid having multiple kinds of user ID (for instance a user ID in the app component and a separate user ID in the core component), as this can be confusing to users. But as I explained above, having user handles that are separate from the user ID is inevitable if we want to support transactional passkey-registration-on-sign-up auth flows.

Fortunately, **handles are scoped to the passkey component**. Server-side, they are only stored in the component, and they are only passed to the app code so that it can forward them to the frontend. We always call this identifier a *user handle*, the term used in the [WebAuthn spec](https://www.w3.org/TR/webauthn-3/#user-handle) and that won’t be mistaken for an ID. Additionally, we never look up a user by its passkey handle, as the handle value is only used by the authenticator itself to distinguish users.

### DIFF
--- a/packages/core/src/components/passkey/registration.ts
+++ b/packages/core/src/components/passkey/registration.ts
@@ -67,7 +67,7 @@ export const startRegistration = mutation({
       handle = await ctx.db
         .query("handles")
         .withIndex("by_userId", (q) => q.eq("userId", userId))
-        .first();
+        .unique();
       const rows = await ctx.db
         .query("passkeys")
         .withIndex("by_userId", (q) => q.eq("userId", userId))
--- a/packages/core/src/components/passkey/registration.ts
+++ b/packages/core/src/components/passkey/registration.ts
@@ -1,5 +1,6 @@
 import { Infer, v } from "convex/values";
 import { mutation, query } from "./_generated/server";
+import { Id } from "./_generated/dataModel";
 import {
   ClientDataType,
   coseAlgorithmES256,
@@ -14,15 +15,22 @@ import {
   finishRegistrationUserError,
   deletePasskeyUserError,
 } from "./validation";
-import { consumeChallenge, randomChallenge, toArrayBuffer } from "./helpers";
+import {
+  consumeChallenge,
+  randomChallenge,
+  randomHandle,
+  toArrayBuffer,
+} from "./helpers";
 import { scheduleChallengeCleanup } from "./cleanup";
 
-// The challenge and the credential IDs travel as raw bytes (Convex
-// `v.bytes()` carries `ArrayBuffer`s end to end). The WebAuthn API in the
-// browser makes and accepts the same bytes, so no base64 conversion is
-// necessary.
+// The challenge, the credential IDs, and the user handle travel as raw bytes
+// (Convex `v.bytes()` carries `ArrayBuffer`s end to end). The WebAuthn API
+// in the browser makes and accepts the same bytes, so no base64 conversion
+// is necessary.
 const startRegistrationResult = v.object({
   challenge: v.bytes(),
+  // The WebAuthn user handle (`user.id`) for the `create()` call.
+  userHandle: v.bytes(),
   excludeCredentials: v.array(v.bytes()),
 });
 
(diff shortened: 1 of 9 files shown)


Nicolapps:
* **#464** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/464?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#497](https://github.com/get-convex/convex-auth/pull/497) <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/497?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#503](https://github.com/get-convex/convex-auth/pull/503) <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/503?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#463** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/463?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#498** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/498?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#491** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/491?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#462** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/462?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#461** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/461?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#483** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/483?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#467** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/467?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#460** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/460?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/460?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#445** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/445?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2 other dependent PRs ([#489](https://github.com/get-convex/convex-auth/pull/489) <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/489?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#490](https://github.com/get-convex/convex-auth/pull/490) <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/490?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `reboot`


<h2></h2>

packages/core/src/components/passkey/registration.ts
```diff
@@ -33,30 +42,64 @@ const startRegistrationResult = v.object({
...
+ *   none. It also returns the existing credential IDs of the user. This is
+ *   used by the authenticator to ensure there isn’t already a passkey that
+ *   exists for the user.
+ * - Give `null` in the new-account flow (the user row does not exist yet).
+ *   The function makes a new handle with no user. `finishRegistration` links
+ *   the handle to the verified user.
+ *
+ * The `userId` argument is required, not optional: the two flows behave
+ * differently, so the caller must state which flow it runs. (Compare with
+ * `startAuthentication`, where the argument is optional.)
```
dowski, line 57: Instead of a required `userId` for both cases, one of which will be `null`, what if there were `startNewUserPasskeyRegistration` and `startExistingUserPasskeyRegistration` functions? The `userId` would only be required in the latter function.

packages/core/src/components/passkey/registration.ts
```diff
@@ -33,30 +42,64 @@ const startRegistrationResult = v.object({
...
-    await scheduleChallengeCleanup(ctx);
+    let handle: { _id: Id<"handles">; handle: ArrayBuffer } | null = null;
     let excludeCredentials: ArrayBuffer[] = [];
-    if (userId !== undefined) {
+    if (userId !== null) {
+      // A user has a maximum of one handle: reuse it when it exists.
+      handle = await ctx.db
+        .query("handles")
+        .withIndex("by_userId", (q) => q.eq("userId", userId))
+        .first();
```
dowski: Maybe use `.unique()` since more than one handle would be an error.

packages/core/src/components/passkey/registration.ts
```diff
@@ -33,30 +42,64 @@ const startRegistrationResult = v.object({
...
+      // one of them authenticate as the other.
+      const collision = await ctx.db
+        .query("handles")
+        .withIndex("by_handle", (q) => q.eq("handle", bytes))
+        .first();
+      if (collision !== null) {
+        throw new Error(
+          "The new user handle collides with an existing handle.",
+        );
+      }
```
dowski, line 92: No change needed, but this is sure the kind of place where a unique constraint would be useful.

dowski (approved):

packages/core/src/components/passkey/registration.ts
Nicolapps: Yes, I fully agree!

packages/core/src/components/passkey/registration.ts
Nicolapps: Thanks, I changed it in 726f87895f549a803871a37c70dbbc245be4d6ca

packages/core/src/components/passkey/registration.ts
Nicolapps: I agree it would be better, I added a TODO comment (b6bea021cbf2b80074a939556fd746dd6d4f3796)

PR #466, opened 2026-08-13
Add a warning that will let React Native apps know they need to configure token storage

dowski:
Also includes some tests that cover using promise based storage in the app.

### DIFF
--- a/packages/core/src/browser/storage.ts
+++ b/packages/core/src/browser/storage.ts
@@ -50,14 +50,34 @@ export class InMemoryStorage implements TokenStorage {
   }
 }
 
+let warnedAboutInMemoryFallback = false;
+
+/** Exposed for tests; the warning is otherwise once per process. */
+export function resetInMemoryFallbackWarning(): void {
+  warnedAboutInMemoryFallback = false;
+}
+
 /**
  * The default storage: the browser's `localStorage` when available, otherwise
  * an {@link InMemoryStorage}. Callers that need cross-tab sessions, React
  * Native, or per-tab sessions should pass their own storage instead.
  */
 export function defaultStorage(): TokenStorage {
-  if (typeof window !== "undefined" && window.localStorage) {
-    return window.localStorage;
+  if (typeof window !== "undefined") {
+    if (window.localStorage) return window.localStorage;
+    // React Native has a window with no `localStorage`. Warn once here that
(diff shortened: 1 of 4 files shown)


dowski:
* **#466** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/466?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/466?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `reboot`


<h2></h2>

packages/core/src/browser/storage.ts
```diff
@@ -50,14 +50,34 @@ export class InMemoryStorage implements TokenStorage {
...
+    if (window.localStorage) return window.localStorage;
+    // React Native has a window with no `localStorage`. Warn once here that
+    // in-memory storage will be used until another storage implementation is
+    // supplied.
+    if (!warnedAboutInMemoryFallback) {
+      warnedAboutInMemoryFallback = true;
+      console.warn(
+        "[convex-auth] This runtime has no `localStorage`, so the session is " +
+          "being kept in memory and will not survive an app restart. Pass a " +
+          "`storage` implementation to your auth provider to persist it." +
```
erquhart: nit: needs a space or the period will run into the next sentence

packages/core/src/browser/storage.ts
```diff
@@ -50,14 +50,34 @@ export class InMemoryStorage implements TokenStorage {
...
+          "For React Native, that would be one backed by `expo-secure-store` " +
+          "or `AsyncStorage`.",
```
erquhart: AsyncStorage isn't suitable for the tokens we're storing, we should just recommend expo-secure-store

erquhart (commented):
Not in this PR but related - #attachStorageListener calls window.addEventListener(), which will throw in RN before anything can really happen if we don't check for it to be defined first.

dowski:
That one should be resolved in https://app.graphite.com/github/pr/get-convex/convex-auth/465/Fix-lifecycle-issue-on-React-Native.

packages/core/src/browser/storage.ts
dowski: Thanks, updated!

erquhart (approved):

PR #469, opened 2026-08-14
Bring back setup functions

dowski:
Bring back setup functions

This allows each provider to have its own callback for handling sign-in/sign-up.

It also makes provider creation somewhat less "magical". Providers are functions that take result of calling `setupCore` which includes custom helpers (`authMutation`, `authAction`) that have access on `ctx` to functions for completing sign in, looking up users, etc. The pattern is inspired by `convex-helpers` and its custom functions pattern.

Two nice benefits:

1. It's easy to scope behavior to providers
2. The callbacks can now be typed directly to the provider's profile type

### DIFF
--- a/examples/react-minimal/convex/users.ts
+++ b/examples/react-minimal/convex/users.ts
@@ -1,23 +1,22 @@
 import { internalMutation } from "./_generated/server";
 import { v } from "convex/values";
 
-export const upsertFromAuth = internalMutation({
+export const createOrUpdateUser = internalMutation({
   args: {
-    provider: v.union(v.literal("anonymous")),
+    provider: v.literal("anonymous"),
     providerAccountId: v.string(),
-    profile: v.any(),
+    profile: v.object({}),
     userId: v.union(v.string(), v.null()),
   },
   returns: v.id("users"),
   handler: async (ctx, args) => {
-    switch (args.provider) {
-      case "anonymous": {
-        return ctx.db.insert("users", {});
+    if (args.userId !== null) {
+      const existing = ctx.db.normalizeId("users", args.userId);
+      if (existing === null) {
+        throw new Error(`Unknown user id: ${args.userId}`);
       }
-      default: {
-        const _exhaustive: never = args.provider;
-        return _exhaustive;
-      }
+      return existing;
     }
+    return ctx.db.insert("users", {});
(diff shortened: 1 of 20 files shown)


dowski:
* **#469** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/469?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/469?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `reboot`


<h2></h2>

examples/react-minimal/convex/users.ts
```diff
@@ -1,23 +1,22 @@
...
 
-export const upsertFromAuth = internalMutation({
+export const createOrUpdateUser = internalMutation({
   args: {
-    provider: v.union(v.literal("anonymous")),
+    provider: v.literal("anonymous"),
     providerAccountId: v.string(),
-    profile: v.any(),
+    profile: v.object({}),
     userId: v.union(v.string(), v.null()),
```
Nicolapps: It would be amazing if we could change this to `v.id("users")`, it would remove the need for the `normalizeId` call

packages/core/src/lib/types.ts
```diff
@@ -182,24 +182,57 @@ export const vCreateOrUpdateUser = v.object({
...
   "internal",
   {
     provider: Provider;
     providerAccountId: string;
-    // TODO: dowski - investigate type safety for provider profile data.
-    profile: unknown;
+    profile: Profile;
     userId: string | null;
   },
-  GenericId<"users">
```
Nicolapps, line 195: Why do we need that change? This might cause users to write callbacks that don't return the right value

Nicolapps (approved):

examples/react-minimal/convex/users.ts
dowski: Done!

packages/core/src/lib/types.ts
dowski: Yeah good pushback. I didn't like the idea of requiring that the table be called "users" but we can allow a custom users table name while still adding the type safety that you suggested.

PR #472, opened 2026-08-15
Password component: reject common passwords

Nicolapps:
This modifies the password component to automatically reject passwords that are among the most commonly used. Doing so improves the security of the application, and is required by security standards (e.g. [OWASP ASVS 6.2.4](https://x.com/rauchg/status/2088456443724955734)).

In order to implement this feature, a list of the 3000 most common passwords is now bundled in the password component. The size of this file is reasonable (43 KiB), and this file is only loaded at runtime when running password component functions.

The file is automatically generated from a script that retrieves the [NCSC top 100k most used passwords](https://github.com/danielmiessler/SecLists/blob/master/Passwords/Common-Credentials/100k-most-used-passwords-NCSC.txt) and extracts the 3000 top passwords. The passwords are grouped by length, normalized, and sorted, which makes finding whether a password is in the list efficient.


Nicolapps:
* **#472** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/472?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/472?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `reboot`


<h2></h2>

packages/core/src/components/password/commonPasswords.generated.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Each password is in the NFC form and is lowercase. The key of the map is the
+ * length of the password in Unicode code points. Each array is in ascending
+ * order, thus `commonPasswords.ts` can do a binary search.
+ *
+ * @generated
+ * @module
+ */
+
+// prettier-ignore
+export const COMMON_PASSWORDS_BY_LENGTH: Record<number, readonly string[]> = {
```
dowski, line 24: It looks like 10 is the shortest password included. Any thoughts on including shorter ones too to offer the protection to apps that don't want to use min 10 char passwords?

dowski: Ah I see you have that as a TODO later. Would it hurt to make the min smaller out of the gate rather than have it be configurable later?

dowski (approved):

packages/core/src/components/password/commonPasswords.generated.ts
Nicolapps: Yeah, I plan to change this, I think I will reduce the min count to 8 but default the provider function to higher minimum lengths by default

PR #474, opened 2026-08-16
Add core signUpWithoutSession: account + user without a session

Nicolapps:
This PR adds a new `createAccount` function to the core component: it is similar to `signIn` as it also resolves the account and create it if necessary, but unlike `signIn` it does not mint a session.

This is useful for the email validation flow, because in this flow we create the `users` row before the email is validated.

I'm not sure whether this is a good API: we'll likely want to revisit the core API boundaries before the GA v2 release (I added a TODO comment).

### DIFF
--- a/packages/core/src/components/core/public.ts
+++ b/packages/core/src/components/core/public.ts
@@ -353,6 +353,36 @@ export const signIn = mutation({
 });
 
 /**
+ * Create the account and the app user for a provider's verified identity
+ * claims, without minting a session.
+ *
+ * Providers use this (via the `createAccount` helper the core hands them)
+ * when the user must complete a step before the first sign-in — for example,
+ * an email validation. Account and user resolution follows the same rules as
+ * `signIn` (the app's `createOrUpdateUser` mutation runs, and
+ * `USE_USER_ID_AS_ACCOUNT_ID` keys the account by the minted user id), but no
+ * refresh token or access token is issued: the user cannot make authenticated
+ * calls until a later `signIn` succeeds.
+ */
+export const createAccount = mutation({
+  args: {
+    claims: vAuthClaims,
+    createOrUpdateUserHandle: v.string(),
+  },
+  returns: v.object({ userId: v.string() }),
+  handler: async (ctx, args): Promise<{ userId: string }> => {
+    // TODO: This API is kind of awkward. We might want to reconsider it before the GA v2 release.
+
+    const { userId } = await resolveAccount(
+      ctx,
+      args.claims,
+      args.createOrUpdateUserHandle as CreateOrUpdateUserFunctionHandle,
+    );
+    return { userId };
+  },
+});
+
+/**
  * Resolve a provider identity to its app user id without minting a session.
  *
  * Providers use this (via the `resolveUserId` helper the core hands them) to look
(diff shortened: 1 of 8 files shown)


Nicolapps:
* **#480** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/480?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#514** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/514?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#479** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/479?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#478** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/478?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#477** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/477?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#476** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/476?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#475** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/475?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#474** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/474?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/474?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `reboot`


<h2></h2>

dowski:
Full review to follow, but re: API, I think the create-user-but-no-session thing is an example of what the requirements/incomplete sign-in system should provide.

What I think this stack demonstrates though is that we're going to want to have some default requirements that are installed by providers. That way we can have the EmailPassword provider require a validated email by default.

packages/core/src/components/core/public.ts
```diff
@@ -276,6 +276,36 @@ export const signIn = mutation({
...
+ *
+ * Providers use this (via the `createAccount` helper the core hands them)
+ * when the user must complete a step before the first sign-in — for example,
+ * an email validation. Account and user resolution follows the same rules as
+ * `signIn` (the app's `createOrUpdateUser` mutation runs, and
+ * `USE_USER_ID_AS_ACCOUNT_ID` keys the account by the minted user id), but no
+ * refresh token or access token is issued: the user cannot make authenticated
+ * calls until a later `signIn` succeeds.
+ */
+export const createAccount = mutation({
```
dowski: In https://app.graphite.com/github/pr/get-convex/convex-auth/488 there's now an internal `createAccount` that might now form the basis of this exposed function.

dowski (approved):
I'm okay proceeding with this even though I think in the end this will fit well with the requirements system. It seems good to be able to push forward and prove out other parts of the email verification subsystem that won't be touched by that future change.

Nicolapps:
I renamed my new function to signUpWithoutSession to avoid the naming conflict with your new function

packages/core/src/components/core/public.ts
Nicolapps: Done, thanks

Nicolapps:
### Merge activity

* **Aug 20, 12:16 AM UTC**: @Nicolapps merged this pull request with [Graphite](https://app.graphite.com/github/pr/get-convex/convex-auth/474).

PR #476, opened 2026-08-16
authEmail: start implementation of email challenges

Nicolapps:
This PR starts the second half of the `authEmail` component: prove that an address belongs to a user.

## What is a challenge?

This PR adds code to create and verify _challenges_, which are emails send to an email address to control that the user has access to it.

Why do we name this a challenge (instead of e.g. a _verification attempt_)? It’s because this machinery will be used both for:
- email validation on account creation (sends an email to a _new address email_)
- account recovery or re-authentication before dangerous actions (sends an email to an email address that is already verified).

So using the term _challenge_ is deliberate here: we are not only going to use this for the initial email verification.

## What does this PR do?

This PR adds code the base implementation of challenges. For now we only provide helper functions that will be used by the component functions. The actual functions to create/verify a challenge are implemented upstack.

## How do purposes work?
Challenges will have multiple purposes:

- **Challenges that have a meaning within the email component itself**
  * These challenges will create a row in the email component’s `verifiedEmails` table. Since `verifiedEmails` has no other way of creating a row, the structure of the component ensures that all rows in `verifiedEmails` are correct.[^1]
  * We have two challenges for now:
    - `setPrimaryEmail`, which is good for apps that assume there is a single email address: on completion, the verified email will fully replace the old email address if there is already one in `verifiedEmails`
    - `addEmail`, which is similar but adds the email address as secondary if there is already one.
- **Other challenges**: they don’t do anything in the component itself, but can be used to implement custom behavior that requires email verification. This can include both:
  - Flows within the auth provider (e.g. account recovery).
  - Custom app-specific flows (e.g. reauthenticate before doing an app-specific action).
  - This PR doesn’t have support for these custom flows. These are implemented upstack (see #540).

[^1]: This means that for now, `verifiedEmails` can only contain emails verified within the email component itself. We might extend this to support 3rd-party verification sources (e.g. OAuth, Convex Auth v1 migration…), but it might require something like component attestation in order to be sound.

## How do rate limits work?

There are two kind of rate limits we need to implement for email verification:
- Challenge creation rate limits (to avoid someone getting spammed by verification attempts started by a malicious user with their email address): these are not implemented yet, but will be necessary. They are left out from this PR but will be added in https://github.com/get-convex/convex-auth/pull/618
- Challenge verification attempt rate limit (to avoid someone trying to bruteforce the right challenge response): for now, this component only issues _long_ challenges, which have high enough entropy, so we don’t need these rate limits.


Nicolapps:
* **#476** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/476?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#530](https://github.com/get-convex/convex-auth/pull/530) <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/530?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#539](https://github.com/get-convex/convex-auth/pull/539) <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/539?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/476?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#528** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/528?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `reboot`


<h2></h2>

packages/core/package.json
```diff
@@ -135,13 +135,21 @@
       "types": "./dist/components/passkey/react.d.ts",
       "default": "./dist/components/passkey/react.js"
     },
-    "./email/*.js": {
```
Nicolapps, line 150: Replacing the wildcard export by manual exports because I expected this component to have a lot of internal-only files, like the passkey component

packages/core/src/components/email/challenge/common.ts
```diff
@@ -0,0 +1,219 @@
...
+ * ```
+ *
+ * The completion step runs when the landing page opens with a link. The page
+ * reads the code from the URL and the secret from storage, then completes at
+ * once:
+ *
+ * ```
+ *  Landing page                Provider                          Component
+ *    │                           │                                 │
+ *    ├─▶ read the code from the URL and the secret from storage    │
```
dowski, line 46: What if we did OTP code instead of anything based on a link?

If we did that, then it mitigates the "attacker gets a user to give access to an account they control" attack vector, which removes the need for two secrets and overall simplifies the verification (e.g. no need for an HTTP route for verification).

dowski (commented):
Sending one high level suggestion before getting into the details of review. Better to cover it first I think, in case we change direction.

packages/core/src/components/email/challenge/common.ts
Nicolapps: As I mentioned in our 1:1, we are eventually going to support both OTP and links. Both are going to use client secrets so that you can’t accidentally click a link and complete a flow started by an attacker.

packages/core/src/components/email/challenge/common.ts
```diff
@@ -0,0 +1,219 @@
...
+
+export type CompleteChallengeFailure = Infer<typeof completeChallengeFailure>;
+
+/**
+ * The result of `claimChallenge`. The kinds return `failure` as-is when the
+ * claim fails; it already has the shape of a failed `complete` result.
+ */
+export type ClaimChallengeResult =
+  | { ok: true; row: Doc<"challenges"> }
+  | { ok: false; failure: CompleteChallengeFailure };
```
dowski: Any reason we're using `ok` here instead of `success` like the other types?

packages/core/src/components/email/challenge/common.ts
```diff
@@ -0,0 +1,219 @@
...
+ * browser keeps, and the ID of the new row.
+ */
+export async function createChallenge(
+  ctx: MutationCtx,
+  args: {
+    email: string;
+    purpose: ChallengePurpose;
+    ttlMs: number;
+  },
+): Promise<{ browserSecret: string; challengeId: Id<"challenges"> }> {
```
dowski, line 152: nit: express this in terms of `StartChallengeResult`.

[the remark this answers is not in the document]
dowski: This seems relevant.

packages/core/src/components/email/challenge/common.ts
```diff
@@ -0,0 +1,219 @@
...
+  args: { emailCode: string; browserSecret: string; purpose: ChallengePurpose },
+): Promise<ClaimChallengeResult> {
+  const browserSecretHash = await sha256Hex(args.browserSecret);
+  const row = await ctx.db
+    .query("challenges")
+    .withIndex("by_browserSecretHash", (q) =>
+      q.eq("browserSecretHash", browserSecretHash),
+    )
+    .unique();
+  if (row === null || row.expiresAt < Date.now()) {
```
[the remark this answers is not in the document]
dowski: It's fine to defer cleanup to one place, but add a TODO if you're going to chase it down later.

packages/core/src/components/email/challenge/common.ts
```diff
@@ -0,0 +1,219 @@
...
+    return claimFailure("INVALID_CHALLENGE");
+  }
+  if (row.emailCodeHash !== (await sha256Hex(args.emailCode))) {
+    return claimFailure("INCORRECT_CODE");
+  }
```
dowski, line 211: What do you think about logging these? It seems like it could be helpful to chase down a bug or misuse.

dowski (approved):

packages/core/src/components/email/challenge/common.ts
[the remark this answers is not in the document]
Nicolapps: Yep good idea, thanks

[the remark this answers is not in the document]
Nicolapps: (I’m ignoring this for this PR since the PR that adds the cleanup is just above in the stack, but I’m going to consider this for future PRs)

packages/core/src/lib/crypto.ts
```diff
@@ -15,3 +15,19 @@ export async function sha256Hex(value: string): Promise<string> {
...
+    .replace(/\+/g, "-")
+    .replace(/\//g, "_")
+    .replace(/=+$/, "");
+}
+
+/**
+ * Cryptographically random 256-bit opaque token, base64url encoded. Used for
+ * the tokens that a challenge or a ticket hands out and later hashes.
+ */
+export function generateRandomToken(): string {
```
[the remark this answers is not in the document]
Nicolapps: That’s fine, it’s a one-line function

packages/core/src/components/email/challenge/common.ts
Nicolapps: Yep, good idea, logged in 032daae6247a8efc2ed43c0632eb2f39b845e153

packages/core/src/components/email/challenge/common.ts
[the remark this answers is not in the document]
Nicolapps: Yeah. Looks like this is fixed upstream by #617 so I'm going to ignore this in this PR

Nicolapps: I don’t think it’s worth it, this is an internal function and it doesn’t structurally need to be identical to the public function return value.

PR #484, opened 2026-08-18
Add provider client setup, storage, and state management

erquhart:
- Accept client provider setup functions in the AuthClient (and ConvexAuthProvider)
- Client setup is two phases:
  - synchronous in AuthClient constructor for setting up state prior to first render
  - optional `onInit` that runs in the client's init before session load for side effect work like processing an oauth callback url
- Each setup gets a store (memory/state) and storage (persistent) view scoped to its id to avoid collisions
- Pull the refresh retry out of the session manager so the oauth client can share it, and fix it to use its last backoff (it slept 2s and then gave up, now it makes a third attempt)


packages/core/src/browser/keyedStore.ts
```diff
@@ -0,0 +1,101 @@
...
+  get<T>(key: string): T | undefined;
+  /**
+   * Subscribe to changes of the value at `key` within the scope. Returns an
+   * unsubscribe function.
+   */
+  subscribe(key: string, listener: () => void): () => void;
+};
+
+/** The full store key for `key` within the scope named `prefix`. */
+export function scopedKey(prefix: string, key: string): string {
```
dowski: It looks like this function is only used internally (except in the test). Probably doesn't need to be exported (or directly tested).

packages/core/src/browser/providerSetup.ts
```diff
@@ -0,0 +1,90 @@
...
+  FunctionReturnType,
+} from "convex/server";
+import type { ScopedKeyedStore } from "./keyedStore";
+import type { AuthClient } from "./sessionManager";
+import type { ScopedStorage } from "./storage";
+
+/**
+ * How a provider's sign-in functions get executed. Provider code takes this
+ * as given, at setup time via {@link AuthProviderClientContext} and in hooks
+ * via `useAuthSignInApi`, instead of picking a transport itself. Executing
```
dowski: Should this refer to hooks since it's meant to be framework agnostic?

packages/core/src/browser/sessionManager.test.ts
```diff
@@ -255,6 +257,75 @@ describe("AuthClient", () => {
     expect(() => client.dispose()).not.toThrow();
   });
 
+  test("withSignInPending reports loading while a completion runs", async () => {
+    const { client } = makeClient();
```
dowski, line 262: Is there any way this can be modeled more like the real use that the setup code module doc describes? E.g. the `onInit` that `withSignInPending` returns is called within `client.init`?

dowski: Oh I see the test below maybe comes closer to the real thing, but maybe not quite there? In any case, are both tests needed? Maybe just one that models the actual intended use.

packages/core/src/browser/sessionManager.ts
```diff
@@ -223,15 +226,72 @@ export class AuthClient {
...
+        throw new Error(
+          `Provider client setup id "${id}" is registered twice; each auth ` +
+            `method registers once per ConvexAuthProvider`,
+        );
+      }
+      seen.add(id);
+      const registration = setup({
+        client: this,
+        store: this.#store.scoped(id),
+        storage: this.#storage.scoped(id),
```
dowski: Seeing `store` and `storage` right here beside each other make me wonder if there's a better name for one or the other.

packages/core/src/browser/storage.ts
```diff
@@ -115,4 +135,19 @@ export class NamespacedStorage {
...
     return this.storage.removeItem(this.key(key));
   }
+
+  /**
+   * A {@link ScopedStorage} for the provider client registered under `id`,
+   * mapping `key` to `__convexAuthProvider_<id>_<key>` before the namespace
+   * suffix. Provider keys can then never collide with the core token keys or
+   * another provider's.
+   */
+  scoped(id: string): ScopedStorage {
```
dowski: Having something called `scoped` inside something called `Namespaced` is a bit confusing. Maybe calling it `providerScoped` would help?

packages/core/src/react/client.tsx
dowski, line 45: I wonder if we should rename `useAuthActions` -> `useAuthClientActions`, since now with `withSignInPending` they may operate on just the client itself.

packages/core/src/react/providers.ts
```diff
@@ -0,0 +1,50 @@
+/**
+ * React building blocks for provider-client authors, exported at
+ * `@convex-dev/auth/react/providers`. Apps don't need anything here. Provider
+ * modules (OAuth, and future auth methods) use these to expose their state as
+ * hooks.
+ *
+ * A provider client registers its actions and flow state in the auth client's
+ * keyed store from its {@link AuthProviderClientSetup} (passed to
+ * `ConvexAuthProvider`'s `providerClients` prop); its hooks read them back
+ * with {@link useAuthClientValue}.
```
dowski: Provider clients are only necessary for auth methods that resolve out-of-band/via redirects, right? Could that be more clear in the naming somehow? Direct auth methods like password/passkey/OTP won't need them, right?

packages/core/src/browser/providerSetup.ts
```diff
@@ -0,0 +1,90 @@
+/**
+ * The contract between the core client and provider clients (auth methods).
```
dowski: Having read through everything, I think the part that I realize now but didn't when reading this is that this only applies to out-of-band/redirect based providers. This won't be used for password/passkey/OTP, right?

If that's the case, can you come up with some better name that scopes it to how it's used?

packages/core/src/browser/sessionManager.ts
erquhart: Yeah I thought about this a few times, one is memory and one is persisted. I'll propose better names.

packages/core/src/browser/providerSetup.ts
erquhart: Yeah considered this, but it felt difficult to completely rule out other use cases. If we want to narrow down and rename to purpose I can do that.

packages/core/src/react/client.tsx
erquhart: Not sure I understand (it was always only operational on the client right?), but can make the change

packages/core/src/browser/storage.ts
erquhart: I actually think namespaced might be the fuzzy part, but renaming to providerScoped can for sure help differentiate

packages/core/src/browser/providerSetup.ts
erquhart: ah will fix

packages/core/src/react/client.tsx
dowski: The other functions call through to Convex server functions to refresh tokens or sign out the user.

packages/core/src/browser/sessionManager.test.ts
erquhart: Kept the first of the two as unit coverage of withSignInPending itself, and added a realistically mocked scenario here https://github.com/get-convex/convex-auth/pull/484/changes#diff-15e0692a5b6600cb57a147781d4611bc3e5fd643cb2a5cf7ff9c708b5a775bbeR488

packages/core/src/react/client.tsx
erquhart: withSignInPending is no longer included here.

packages/core/src/react/providers.ts
erquhart: The class of providers that would use this are any that might act on their own, sometimes in response to a url query param (oauth, magic link), but also for device auth and one tap, which isn't redirect or out of band per se. The best term I could come up with for this is "ambient". Renamed things accordingly.

packages/core/src/browser/providerSetup.ts
erquhart: Used prefix "ambient" for this part of the api.

packages/core/src/browser/ambientSignInClient.ts
```diff
@@ -60,30 +67,33 @@ export interface AuthSignInApi {
...
 
 /**
- * A provider client's registration, passed via `ConvexAuthProvider`'s
- * `providerClients` prop and run by the `AuthClient` while it is
+ * An ambient sign-in's registration, passed via `ConvexAuthProvider`'s
+ * `ambientSignIns` prop and run by the `AuthClient` while it is
  * constructed. See the module docs for the two-phase lifecycle.
  */
-export type AuthProviderClientSetup = {
+export type AmbientSignInClient = {
```
dowski, line 90: Would `AmbientSignInHandler` or `AmbientSignInConsumer` work as a name? The "client" term feels pretty overloaded already.

"Client" also feels like something that could be an initiator of an action, and this is more like the receiving end of something that was already initiated, right?

packages/core/src/browser/sessionManager.test.ts
```diff
@@ -354,58 +354,58 @@ const SIGN_IN_API = {
 /** A reference to hand the stub. Its path is never resolved. */
 const SIGN_IN_REF = makeFunctionReference<"mutation">("auth:probeSignIn");
 
-/** An {@link AuthClient} constructed with the given provider client setups. */
-function makeClientWithSetups(
-  setups: ReadonlyArray<AuthProviderClientSetup>,
+/** An {@link AuthClient} constructed with the given ambient sign-ins. */
+function makeClientWithSignIns(
+  signIns: ReadonlyArray<AmbientSignInClient>,
```
dowski, line 359: If `AmbientSignInHandler` works as a name then these would be called `handlers`: `signIns` sounds like data.

dowski: Same with this function name, and a variety of actual implementation names that I think get more understandable  (e.g. `ambientSignIns.handlers` in the `sessionManager` vs. `ambientSignIns.signIns`).

packages/core/src/browser/sessionManager.ts
```diff
@@ -226,55 +225,55 @@ export class AuthClient {
     return this.#accessToken;
   }
 
-  // --- Provider client surface ----------------------------------------------
+  // --- Ambient sign-in surface ----------------------------------------------
```
dowski, line 228: I know I've been picking on naming in this PR, but this is definitely an improvement! This is new mode of operation for the client code and "ambient sign-ins" captures it pretty well.

packages/core/src/browser/storage.ts
```diff
@@ -83,14 +83,14 @@ export function defaultStorage(): TokenStorage {
 }
 
 /**
- * A read/write view over a {@link NamespacedStorage} for one provider
- * client, with every key prefixed by the provider's setup id. The shape
- * handed to provider client setups.
+ * A read/write view over a {@link NamespacedStorage} for one ambient sign-in,
+ * with every key prefixed by that sign-in's id. The shape handed to an
+ * ambient sign-in's setup.
```
dowski, line 88: Maybe include a note here re: the purpose of `SignInStorage`. As I understand it, it allows the client to stash some values that will be important for it to be able to retrieve later when handling an ambient sign-in result from something else that helps complete the sign-in.

packages/core/src/browser/ambientSignInClient.ts
```diff
@@ -0,0 +1,99 @@
+/**
+ * The contract between the core client and ambient sign-ins.
+ *
+ * Ambient sign-ins run on their own at client startup, not because the app
```
dowski, line 4: Rewording suggestion for the first sentence of this paragraph.

> Ambient sign-ins run in their own handler(s) when the client code initializes. That enables them to intercept ambient sign-in data (e.g. from the URL) to complete an earlier sign-in that the client code initiated.

packages/core/src/browser/ambientSignInClient.ts
```diff
@@ -0,0 +1,99 @@
...
+ * called something. OAuth is the case today: the user comes back from Google
+ * on some page that never called a sign-in hook, and the client has to notice
+ * and finish it. They are registered whether the app uses them or not, so the
+ * startup code has to return right away unless it finds a sign-in of its own
+ * to finish.
```
dowski, line 9: Rewording last sentence here:

> Ambient sign-in handlers are always active once registered, and should run as little logic as needed and return early if there is no pending sign-in to complete.

packages/core/src/browser/ambientSignInClient.ts
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * 1. **Setup**: runs synchronously while the `AuthClient` is constructed.
+ *    Register actions and set starting values in `ctx.values` here, so they
+ *    exist before the app's first render reads them.
+ * 2. **Init**: the optional returned `onInit` runs inside the client's
+ *    `init()`, once per client instance, before the persisted session
+ *    loads. Side-effectful work belongs here (e.g. handling an OAuth
+ *    callback URL). Call `withSignInPending` synchronously in `onInit`
+ *    around a sign-in completion, so the auth state reports loading rather
+ *    than briefly signed out.
```
dowski, line 27: More rewording:

> When an `AmbientSignInHandler` decides it needs to complete a sign-in, it should call `withSignInPending` synchronously in `onInit` so the auth state reports loading rather than briefly signed out

packages/core/src/browser/ambientSignInClient.ts
```diff
@@ -0,0 +1,99 @@
...
+ * called something. OAuth is the case today: the user comes back from Google
+ * on some page that never called a sign-in hook, and the client has to notice
+ * and finish it. They are registered whether the app uses them or not, so the
+ * startup code has to return right away unless it finds a sign-in of its own
+ * to finish.
+ *
+ * A provider module exports a factory (e.g. `oauth(...)`) producing an
+ * {@link AmbientSignInClient} the app passes to `ConvexAuthProvider`'s
+ * `ambientSignIns` prop. Each registration carries a unique alphanumeric
+ * `id` naming the auth method. The values and storage views the setup
```
dowski, line 14: Can you explain more what "the setup" is here? It's possible this part should just be reworked.

packages/core/src/browser/ambientSignInClient.ts
dowski: Or `PendingSignInHandler` (got this idea after reading about the interaction with`withSignInPending` above).

packages/core/src/browser/keyedStore.ts
```diff
@@ -0,0 +1,101 @@
+/**
+ * A keyed, subscribable value store scoped to one `AuthClient` instance.
+ *
+ * Ambient sign-ins publish here what their hooks need to read (the actions
```
dowski, line 4: nit: hooks are React-specific.

dowski (approved):
This is good, thanks for bearing with all of my naming suggestions! I feel like I understand the code better now thanks to what you've done, which is much of the battle these days.

One thing I'm still not 100% clear on is the relationship between `SignInValues`/`KeyedStorage`/`SignInStorage`/`NamedSpacedStorage`.

I *think* the first couple (`SignInValues/KeyedStorage`) are more transient and are effectively for communication between components.

The the latter are used for more *persistent* values that need to survive redirects and things.

Is that the right understanding?

----

Anyhow, take whichever of my additional comments you feel like improves things and ignore any that you think are wrong/too nit-picky.

erquhart:
@dowski I'm going to take another run at naming in general here and the remaining PR comments, merging for now just to keep things moving.

PR #485, opened 2026-08-18
Add oauth provider client

erquhart:
- Handles oauth flow from initialization to completion
- Stores server provided state in its scoped storage and includes it in ticket claim
- Handles code/error for url query params on load
- Basic error reporting with default messaging

### DIFF
--- a/packages/core/src/browser/ambientSignInClient.ts
+++ b/packages/core/src/browser/ambientSignInClient.ts
@@ -26,6 +26,14 @@
  *    around a sign-in completion, so the auth state reports loading rather
  *    than briefly signed out.
  *
+ * Bindings may register setups the app didn't explicitly opt into (the React
+ * binding registers `oauth()` by default), so `onInit` work must be
+ * self-gating. It may only do expensive work (a storage read, a network
+ * call) when it finds local evidence that a flow it owns is actually in
+ * progress, such as a namespaced URL param or a storage key the setup itself
+ * wrote. Absent that evidence it must return cheaply, so an app that never
+ * uses the provider pays nothing at startup.
+ *
  * The Convex imports are type-only. Like `SpaAuthApi`, this contract keeps
  * ambient sign-in logic free of any particular Convex client class, so
  * non-React bindings can drive the same setups.
--- a/packages/core/src/browser/ambientSignInClient.ts
+++ b/packages/core/src/browser/ambientSignInClient.ts
@@ -26,13 +26,9 @@
  *    around a sign-in completion, so the auth state reports loading rather
  *    than briefly signed out.
  *
- * Bindings may register setups the app didn't explicitly opt into (the React
- * binding registers `oauth()` by default), so `onInit` work must be
- * self-gating. It may only do expensive work (a storage read, a network
- * call) when it finds local evidence that a flow it owns is actually in
- * progress, such as a namespaced URL param or a storage key the setup itself
- * wrote. Absent that evidence it must return cheaply, so an app that never
- * uses the provider pays nothing at startup.
+ * `onInit` must gate its own work. A binding can register a setup the app
+ * never asked for, so `onInit` should return right away unless it finds a
+ * sign of its own flow, like a URL param or a storage key it wrote itself.
  *
  * The Convex imports are type-only. Like `SpaAuthApi`, this contract keeps
  * ambient sign-in logic free of any particular Convex client class, so
--- a/packages/core/src/oauth/client.test.ts
+++ b/packages/core/src/oauth/client.test.ts
@@ -1,0 +1,394 @@
+// @vitest-environment jsdom
+import { afterEach, describe, expect, test, vi } from "vitest";
+import { getFunctionName, makeFunctionReference } from "convex/server";
+import { scopedKey } from "../browser/keyedStore";
+import {
+  registerProviderClientSetups,
+  type AuthSignInApi,
+} from "../browser/providerSetup";
+import { AuthClient } from "../browser/sessionManager";
+import { InMemoryStorage, NamespacedStorage } from "../browser/storage";
+import type { TokenBundle } from "../lib/types";
+import {
+  OAUTH_ACTIONS_KEY,
+  OAUTH_FLOW_ERROR_KEY,
+  OAUTH_SETUP_ID,
+  oauth,
+  type OauthActions,
+  type OauthFlowError,
+  type OauthProviderRefs,
+} from "./client";
+
+const NAMESPACE = "https://happy-animal-123.convex.cloud";
+
+const bundle: TokenBundle = {
+  accessToken: "access-1",
+  accessTokenExpiresAt: 0,
+  refreshToken: "refresh-1",
+  refreshTokenExpiresAt: 0,
+  userId: "user-1",
+};
+
+// Real function references (they carry their path), the way an app picks them
+// off its generated `api.auth`. The start leg passes `startSignIn` through by
+// identity; the completion leg rebuilds `completeSignIn` from the persisted
+// path, so completion assertions compare paths via `getFunctionName`.
+const googleStart = makeFunctionReference<"mutation">("auth:startSignInGoogle");
+const googleComplete = makeFunctionReference<"mutation">(
+  "auth:completeSignInGoogle",
+);
+const GOOGLE_REFS: OauthProviderRefs = {
+  providerName: "google",
+  startSignIn: googleStart,
+  completeSignIn: googleComplete,
+};
+
+/** The oauth setup's scoped storage view over `storage`. */
+function flowStorage(storage: InMemoryStorage) {
+  return new NamespacedStorage(storage, NAMESPACE).scoped(OAUTH_SETUP_ID);
+}
+
+/** Store a pending flow the way `signIn` would before navigating away. */
+function seedPendingFlow(
+  storage: InMemoryStorage,
+  {
+    providerName = "google",
+    state = "state-1",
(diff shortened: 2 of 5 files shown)


packages/core/src/browser/ambientSignInClient.ts
```diff
@@ -19,6 +19,10 @@
  *    around a sign-in completion, so the auth state reports loading rather
  *    than briefly signed out.
  *
+ * `onInit` must gate its own work. A binding can register a setup the app
+ * never asked for, so `onInit` should return right away unless it finds a
```
dowski, line 30: What does

> A binding can register a setup the app never asked for

mean?

packages/core/src/oauth/client.test.ts
```diff
@@ -0,0 +1,397 @@
...
+function calledPath(mutation: ReturnType<typeof vi.fn>, call = 0): string {
+  return getFunctionName(mutation.mock.calls[call]![0] as never);
+}
+
+describe("OAuth client", () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+    window.history.replaceState(null, "", "/");
+    // Restore the prototype getter shadowed by the React Native stub.
+    delete (window.navigator as { product?: string }).product;
```
dowski: We have some other tests that restore `window` entirely. That might be better in case there's an initial value for `product` that will never get restored.

packages/core/src/oauth/client.ts
```diff
@@ -0,0 +1,316 @@
...
+
+/** Why a sign-in failed, with default copy an app can render as-is. */
+export type OauthFlowError = {
+  /** Why the sign-in failed. */
+  code: OauthFlowErrorCode;
+  /**
+   * Default English copy for `code`. Not a stable string. To localize or
+   * rebrand, switch on `code` and ignore this.
+   */
+  message: string;
```
dowski: Maybe just leave this out? Our password and passkey providers just use codes so that would be consistent.

packages/core/src/oauth/client.ts
```diff
@@ -0,0 +1,316 @@
...
+  state: string;
+  /**
+   * The path of the provider's `completeSignIn` mutation, from
+   * `getFunctionName`. The flow can return to any page of the app, including
+   * one that never held the mutation reference, so the path is saved here and
+   * the reference is rebuilt from it. If the app renamed that export and
+   * redeployed mid-flight the path no longer resolves and the sign-in fails
+   * as `oauth_error`.
+   */
+  completeSignIn: string;
```
dowski, line 138: Can we avoid this somehow?

packages/core/src/browser/ambientSignInClient.ts
erquhart: The general idea is that we want provider client init to be able to run invisibly when its not needed. The configure-once concept means server config without having to match up a client config, so a client provider might have default provider clients registered that the app doesn't actually use, which is okay if those clients are guaranteed to do nothing. If the presence of a client provider means stuff might happen then the app has to opt in to every provider client and we can't provide any by default.

packages/core/src/oauth/client.ts
erquhart: The goal here is to make it turnkey and provide working defaults (the average app won't have big opinions on what message to use, nice to just have one there), but I can strip out if preferred.

packages/core/src/browser/ambientSignInClient.ts
dowski: Oh, so you mean we might register the OAuth callback handling code by default but it will be latent until it's activated by the app deciding to hook up an OAuth provider?

packages/core/src/oauth/client.ts
dowski: I like the idea you mentioned offline of having defaults at a higher level.

packages/core/src/browser/ambientSignInClient.ts
erquhart: Yes but since it doesn't know for sure if you hooked up an oauth provider on the server side, its latent until its own check passes, which is the presence of a convexAuthCode or convexAuthError query param in the url when app starts

packages/core/src/oauth/client.ts
erquhart: If you mean the mid-flight error, its pretty theoretical and the outcome is arguably best case. If an app developer who has already deployed convex auth and has users on it decides to rename the completeSignIn mutation to something else, anyone attempting to sign in at the moment of deployment would just get an opaque error with no other side effect. They could then try to sign in again and it would work.

erquhart: so basically it will be rare if it happens and if it does it likely won't happen more than once in an app's lifetime. and the ux of a failed sign in probably won't register since they'll fail exactly once, re-attempt, and they're signed in

packages/core/src/oauth/client.test.ts
erquhart: This test file runs in the jsdom env (`// @vitest-environment jsdom` at top), so we can't just mock/restore the window, but I improved how we handle this value.

packages/core/src/oauth/client.ts
```diff
@@ -0,0 +1,300 @@
...
+export type OauthActions = {
+  /**
+   * Start the given provider's OAuth flow, or finish a saved one when
+   * `options.code` is set. Starting navigates away to the identity provider.
+   */
+  signIn: (
+    refs: OauthProviderRefs,
+    options?: SignInOptions,
+  ) => Promise<SignInOutcome>;
+};
```
dowski, line 99: Could this be split into functions to start and finish, rather than that being interpreted via which options are passed?

packages/core/src/oauth/client.ts
dowski: So do you think it's worth having the functionality in place to handle that scenario? It seems like a user would get some error in either case and likely retry signing in. My preference is to not have code to handle that.

packages/core/src/oauth/client.ts
```diff
@@ -0,0 +1,300 @@
...
+
+/**
+ * Client setup for the OAuth providers. It owns the flow's storage and the
+ * startup work that finishes a flow. Provider mutations arrive with each
+ * {@link OauthActions.signIn} call, so the setup takes no configuration.
+ */
+export function oauth(): AmbientSignInClient {
+  const setup: AmbientSignInClient["setup"] = ({
+    client,
+    values,
```
dowski, line 185: `values` here is basically the API for interacting with OAuth, right? Would it make sense to call it `oAuthApi` or something? Or is it too low-level to call it an API? `oAuthDataChannel`? Anything with `state` probably doesn't work b/c that gets confusing w/ the state used in the handshake. Hmm.

packages/core/src/oauth/client.ts
```diff
@@ -0,0 +1,300 @@
...
+  }
+}
+
+/**
+ * Client setup for the OAuth providers. It owns the flow's storage and the
+ * startup work that finishes a flow. Provider mutations arrive with each
+ * {@link OauthActions.signIn} call, so the setup takes no configuration.
+ */
+export function oauth(): AmbientSignInClient {
+  const setup: AmbientSignInClient["setup"] = ({
```
dowski, line 183: There's a few layers nested functions defined here - could these be broken up at all to make it more clear?

packages/core/src/oauth/client.test.ts
```diff
@@ -0,0 +1,407 @@
...
+    mode: "spa",
+    authApi: { refreshSession: async () => null, signOut: async () => {} },
+    storage,
+    storageNamespace: NAMESPACE,
+    ambientSignIns: { signIns: [oauth()], signInApi },
+  });
+  const oauthValues = client.ambientSignInValues(OAUTH_SETUP_ID);
+  const actions = oauthValues.get<OauthActions>(OAUTH_ACTIONS_KEY)!;
+  const flowError = () =>
+    oauthValues.get<OauthFlowError | null>(OAUTH_FLOW_ERROR_KEY);
```
dowski, line 75: Seeing this in use, I wonder if `SignInValues` could be more of a structured object instead of a generic values store/channel.

dowski (approved):
My comments here aren't blocking, but curious what you think and if any of them make sense/are easily resolved.

packages/core/src/oauth/client.ts
dowski: Actually `SignInValues` isn't OAuth specific, I guess.

packages/core/src/oauth/client.ts
erquhart: yeah the term "signIn" is overloaded and unhelpful, will address this

packages/core/src/oauth/client.ts
erquhart: I almost think we're saying the same thing? I'm saying we don't have special case logic for it, and you're saying we shouldn't have special case logic for it. Let me know if I'm missing something here, but status is the user would get an error and retry, no special handling is in place.

packages/core/src/oauth/client.ts
erquhart: "signIn" is officially terrible

packages/core/src/oauth/client.ts
erquhart: will circle back and move the nested functions within setup up a level

erquhart:
Will circle back on remaining feedback here.

packages/core/src/oauth/client.ts
dowski: Haha what I was saying is that we shouldn't have to store the string name of the completeSignIn function.

PR #488, opened 2026-08-18
Split the user callback to differentiate new and returning users

dowski:
Split the user callback to differentiate new and returning users

The provider APIs now take a createUser function and an optional onSignIn function

createUser is called only the first time a user signs in. onSignIn is called on every sign in, including the first.

The bulk of the changes are in `core/public.ts` and `core/setup.ts`, but lots of other files were touched in smaller ways (mostly docs/naming). The examples are all updated and show how the API is used. The Next.js one shows how `onSignIn` might be used across providers and the others show that it is optional and doesn't need to be provided.

### DIFF
--- a/examples/nextjs/convex/users.ts
+++ b/examples/nextjs/convex/users.ts
@@ -14,7 +14,7 @@
  */
 import { v } from "convex/values";
 import { Id } from "./_generated/dataModel";
-import { internalMutation, MutationCtx, query } from "./_generated/server";
+import { internalMutation, query } from "./_generated/server";
 
 export const createUserAnonymous = internalMutation({
   args: {
@@ -40,14 +40,9 @@ export const createUserPassword = internalMutation({
   },
 });
 
-/**
- * The per-sign-in hooks. Because `onSignIn` runs on a first sign-in too, right
- * after `createUser`, stamping `lastSignedInAt` here covers every sign-in and
- * the create callbacks don't have to repeat it.
- *
- * One mutation per provider, since each declares its provider's exact literal
- * types, both delegating to a plain shared function.
- */
+// Below is an exmple of using the per-sign-in hooks. Here the app is choosing to
+// record a `lastSignedInAt` timestamp for each user upon sign-in.
+
 export const onSignInAnonymous = internalMutation({
   args: {
     provider: v.literal("anonymous"),
@@ -57,7 +52,7 @@ export const onSignInAnonymous = internalMutation({
   },
   returns: v.null(),
   handler: async (ctx, args) => {
-    return await touchLastSignedIn(ctx, args.userId);
+    await ctx.db.patch("users", args.userId, { lastSignedInAt: Date.now() });
   },
 });
 
@@ -70,17 +65,9 @@ export const onSignInPassword = internalMutation({
   },
   returns: v.null(),
   handler: async (ctx, args) => {
-    return await touchLastSignedIn(ctx, args.userId);
+    await ctx.db.patch("users", args.userId, { lastSignedInAt: Date.now() });
   },
 });
-
-async function touchLastSignedIn(
-  ctx: MutationCtx,
-  userId: Id<"users">,
-): Promise<null> {
-  await ctx.db.patch("users", userId, { lastSignedInAt: Date.now() });
-  return null;
-}
 
 /**
  * The currently signed-in user, or null. Demonstrates an authenticated query
--- a/examples/nextjs/convex/users.ts
+++ b/examples/nextjs/convex/users.ts
@@ -1,10 +1,10 @@
 /**
  * Each provider configured in `convex/auth.ts` has app-owned callbacks defined here.
  *
- * A provider calls `onSignUp` the first time it sees an account, which creates the
- * app's user row and returns its id, and the optional `onSignIn` whenever a known
- * account signs back in. Each callback has a signature that matches the associated
- * provider and the data that it supplies.
+ * A provider calls `createUser` the first time it sees an account, which creates the
+ * app's user row and returns its id, and then calls the optional `onSignIn` on every
+ * sign-in, that first one included. Each callback has a signature that matches the
+ * associated provider and the data that it supplies.
  *
  * In this example, the username and password provider includes the `username` in the
  * profile data that it provides, while the anonymous provider doesn't. In the app's
@@ -16,7 +16,7 @@ import { v } from "convex/values";
 import { Id } from "./_generated/dataModel";
 import { internalMutation, MutationCtx, query } from "./_generated/server";
 
-export const onSignUpAnonymous = internalMutation({
+export const createUserAnonymous = internalMutation({
   args: {
     provider: v.literal("anonymous"),
     providerAccountId: v.string(),
@@ -24,11 +24,11 @@ export const onSignUpAnonymous = internalMutation({
   },
   returns: v.id("users"),
   handler: async (ctx) => {
-    return await createUser(ctx, {});
+    return await ctx.db.insert("users", {});
   },
 });
 
-export const onSignUpPassword = internalMutation({
+export const createUserPassword = internalMutation({
   args: {
     provider: v.literal("password"),
     providerAccountId: v.string(),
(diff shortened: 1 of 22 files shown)


dowski:
* **#492** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/492?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#488** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/488?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/488?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `reboot`


<h2></h2>

examples/nextjs/convex/users.ts
```diff
@@ -14,51 +16,70 @@ import { v } from "convex/values";
...
+export const onSignInAnonymous = internalMutation({
+  args: {
+    provider: v.literal("anonymous"),
+    providerAccountId: v.string(),
+    profile: v.object({}),
+    userId: v.id("users"),
+  },
+  returns: v.null(),
+  handler: async (ctx, args) => {
+    return await touchLastSignedIn(ctx, args.userId);
```
erquhart: This really feels like an internal thing that shouldn't have to happen in user space. It also introduces a field we're requiring the developer to maintain, which we've avoided until now.

erquhart (commented):
I'm a bit unclear on the goals/benefits of the change. What problem(s) does it solve?

dowski:
> I'm a bit unclear on the goals/benefits of the change. What problem(s) does it solve?

Good question. I guess the main benefit is keeping sign-in and sign-up type logic separate.

If an app just wants to run logic on sign-up to populate their user model, they can just use `createUser`. `onSignIn` is optional, but can be used to run logic each time an already created user signs in.

So instead of a `createOrUpdateUser` function that's used for both and you decide whether it's for sign-in or sign-up based on the presence/absence of a passed in `userId`, `createUser` should always create a user when called, and `onSignIn` will always be passed an actual `userId`.

examples/nextjs/convex/users.ts
dowski: Ah this is just an example of something that an app *could* choose to track on each sign-in. Maybe I can come up with a better example.

packages/core/src/components/core/public.ts
```diff
@@ -242,40 +236,122 @@ async function resolveAccount(
...
+ * Providers don't typically call this API directly, but instead use the
+ * framework's `completeSignUp` helper. That function takes care of passing the
+ * `createUserHandle` and `onSignInHandle` app callbacks.
+ *
+ * `createUser` mints the app user, then `onSignIn` runs like it does for any
+ * other sign-in. An `onSignIn` that throws therefore rolls back the user
+ * `createUser` just made, since both are subtransactions of this one.
+ *
+ * The JWT accessToken in the return value is issued with `issuer` as its
+ * `iss`. Token lifetimes default to 1m (access) and 30d (refresh) unless
```
erquhart, line 275: Pre-existing for this pr but we should consider raising the default to at least 5 minutes, I'd have expected 10 or 15.

packages/core/src/components/core/public.ts
```diff
@@ -242,40 +236,122 @@ async function resolveAccount(
...
+ * Providers don't typically call this API directly, but instead use the
+ * framework's `completeSignIn` helper. That function takes care of passing the
+ * `onSignInHandle` app callback.
+ *
+ * The JWT accessToken in the return value is issued with `issuer` as its
+ * `iss`. Token lifetimes default to 1m (access) and 30d (refresh) unless
+ * `accessTokenTtlSeconds` / `refreshTokenTtlSeconds` are supplied (the app
+ * sets these once via `setupCore`).
+ *
+ * Throws when the identity has no account.
```
erquhart, line 318: This repeats most of the signup docstring verbatim

packages/docs/content/docs/login-providers/anonymous.mdx
```diff
@@ -73,6 +73,14 @@ const core = setupCore({ component: components.auth, usersTable: "members" });
 
 ### Set up user creation
 
+Write a `createUser` mutation that creates a row in your `users` table and
+returns its id. The provider calls it for every anonymous sign-in, since each
+one establishes a new account.
+
+You can also attach an optional `onSignIn` mutation, called with the resolved
+`userId` right afterwards. It is the same hook every other provider gets, so
+work an app does on any sign-in has one home.
```
erquhart: "so work an app" I think we need to reword

examples/nextjs/convex/users.ts
erquhart: On reflection this might have just been me missing the point, but maybe clarifying that this itself is an example use case rather than how-to-use-convex-auth would help.

examples/nextjs/convex/users.ts
```diff
@@ -14,51 +16,70 @@ import { v } from "convex/values";
...
+    provider: v.literal("password"),
+    providerAccountId: v.string(),
+    profile: v.object({ username: v.string() }),
+    userId: v.id("users"),
+  },
+  returns: v.null(),
+  handler: async (ctx, args) => {
+    return await touchLastSignedIn(ctx, args.userId);
+  },
+});
```
erquhart, line 70: The amount of wiring we're adding up is making me miss thick clients lol. Will be good to find a way to avoid all of this being in user space outside of advanced cases.

packages/core/src/components/core/public.ts
```diff
@@ -242,40 +236,122 @@ async function resolveAccount(
...
   handler: async (ctx, args): Promise<TokenBundle> => {
     const ttl = resolveTtlConfig(args);
-    const { accountId, userId } = await resolveAccount(
+    const { accountId, userId } = await createAccount(
       ctx,
       args.claims,
-      args.createOrUpdateUserHandle as CreateOrUpdateUserFunctionHandle,
+      args.createUserHandle as CreateUserFunctionHandle,
     );
+    await notifySignIn(ctx, args.claims, userId, args.onSignInHandle);
```
erquhart, line 299: args.claims.providerAccountId would be "" here for password provider, we should patch in the user id. I think there's already a todo for this sort of thing under the previous unified handler so this may already be known.

packages/core/src/lib/types.ts
```diff
@@ -240,13 +281,33 @@ export type CreateOrUpdateUserFn<
...
+   *
+   * Call this when the provider has just created the account. The core records
+   * the account, calls the app's `createUser` to mint the app user, then its
+   * `onSignIn` like any other sign-in, and returns the tokens a client needs to
+   * make authenticated calls.
+   *
+   * Throws if the identity already has an account. A provider that cannot tell
+   * a first sign-in from a return visit should call
+   * {@link BoundAuthHelpers.resolveUserId} first and pick the right helper;
+   * both run in one mutation transaction, so the check cannot go stale.
```
erquhart: Would this potentially not be true in authAction?

erquhart (approved):
Some comments/findings but nothing blocking.

examples/nextjs/convex/users.ts
dowski: Good call, done.

examples/nextjs/convex/users.ts
dowski: Yeah, unfortunately some of this is solely because `FunctionReference` is covariant. Otherwise we could have one function that is more broad in what it accepts and could be used for any of the providers.

Maybe I ought to push more on resolving that.

packages/core/src/components/core/public.ts
dowski: Yeah there's a TODO in `password/setup.ts`. Should definitely revisit.

packages/core/src/components/core/public.ts
dowski: Yep, they do similar things. Should still be useful to have the docs on each one.

packages/core/src/lib/types.ts
dowski: Good callout. I removed that false assurance, and I should probably just remove `authAction` unless we have a use case for it in the framework.

packages/docs/content/docs/login-providers/anonymous.mdx
dowski: Yeah, that was awkward wording. I think this section isn't really useful for the anonymous provider, so I removed it.

PR #500, opened 2026-08-20
Add oauth React hooks

erquhart:
- Per-provider hooks (`useSignInWithGoogle`, `useSignInWithGithub`)
- `ConvexAuthProvider` registers `oauth()` by default, no-op if not used

### DIFF
--- a/packages/core/src/oauth/client.ts
+++ b/packages/core/src/oauth/client.ts
@@ -69,7 +69,8 @@ export type OauthFlowError = {
 /** Options accepted by {@link OauthActions.signIn}. */
 export type SignInOptions = {
   /**
-   * Where the flow returns to when it finishes. Defaults to the current URL.
+   * Where the flow returns to when it finishes. Defaults to the current URL,
+   * and is required where there is no current URL, such as React Native.
    * Must be an http or https URL. Custom schemes like `myapp://` are not
    * supported yet.
    */
@@ -122,10 +123,13 @@ const SERVER_ERRORS: ReadonlySet<string> = new Set([
 const OAUTH_FLOW_STORAGE_KEY = "flow";
 
 /** What `signIn` saves before it navigates to the identity provider. */
-type PendingFlow = {
+export type PendingFlow = {
   /** Which provider the flow belongs to. */
   providerName: string;
-  /** The state minted at sign-in. Proof this client started the flow. */
+  /**
+   * The state the server gave back at sign-in. Proof this client started the
+   * flow.
+   */
   state: string;
   /**
    * The path of the provider's `completeSignIn` mutation, from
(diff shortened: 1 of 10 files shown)


packages/core/src/oauth/client.ts
```diff
@@ -268,12 +285,23 @@ export function oauth(): AmbientSignInClient {
...
       if (options?.code !== undefined) {
+        setFlowError(null);
         return { signedIn: await completeFlow(options.code) };
       }
+      const href = currentHref();
+      const redirectTo = options?.redirectTo ?? href;
+      if (redirectTo === null) {
+        throw new Error(
+          "`redirectTo` is required where there is no current page URL, " +
+            "such as React Native.",
```
dowski, line 297: Partly for my curiosity, what kind of `redirectTo` value would a React Native app set? IIRC we don't support custom schemes for redirects, or maybe that's for the OAuth provider redirect that's handled by the `*.convex.site` code?

packages/core/src/oauth/react.ts
```diff
@@ -0,0 +1,171 @@
...
+ *
+ * OAuth is registered by default in `ConvexAuthProvider`. Each supported
+ * provider ships a hook that reads its sign-in functions from the module you
+ * pass in, usually the generated `api.auth`. {@link useOauth} returns the
+ * state that isn't tied to one provider.
+ *
+ * ```tsx
+ * const { signInGoogle } = useSignInWithGoogle(api.auth);
+ * const { flowError } = useOauth();
+ * await signInGoogle();
```
dowski, line 13: Could the API package up the sign-in function and the flow-error state in one hook instead of requiring a client app to call both?

dowski: Ah I read the docs later, I guess it can be useful to have them split. Maybe `useOauth` could be `useOauthState` or something?

packages/core/src/oauth/react.ts
```diff
@@ -0,0 +1,171 @@
...
+}
+
+/** What {@link useOauthSignIn} returns. */
+export type UseOauthSignInReturn = {
+  /**
+   * Start the provider's OAuth flow (or, with `options.code`, complete one
+   * started elsewhere). Navigates away to the identity provider on the web.
+   * React Native isn't supported yet. It gets the `redirect` URL back to open
+   * in an in-app browser, but `options.redirectTo` is required there and can
+   * only be an http or https URL (see {@link SignInOptions}).
```
dowski, line 91: Ah, I see. I was confused earlier about some code that looked like it was meant to support RN.

dowski (approved):

packages/core/src/oauth/client.ts
erquhart: have to return the code via query param, on web we use whatever url you were on when you attempted sign in but there is no url for RN so you have to manually provide one. But don't put too much stock in any of the RN preparatory stuff, the story is incomplete and I think we should avoid universal link requirements entirely and take advantage of the live websocket in the running app (RN doesn't actually navigate away from the sign in view for oauth, so it can just sit there and listen for a complete sign-in ticket to claim).

May not work for all setups though, like if the sign in link switches apps they still need universal to switch back or tell the user to do so, which is annoying. Need to research a bit more.

PR #505, opened 2026-08-20
Use SimpleWebAuthn for the passkey provider

Nicolapps:
This PR updates the implementation of the passkey component functions to use SimpleWebauthn instead of vendored Oslo dependencies.

Right now, there is no change on the client or in the wire format. This will be done later. Keeping the same wire format allows us to keep using the same tests for now, which also gives us some confidence about the new implementation.


Nicolapps:
* **#580** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/580?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#570** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/570?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#505** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/505?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/505?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `reboot`


<h2></h2>

packages/core/src/components/passkey/authentication.ts
```diff
@@ -270,32 +265,63 @@ export const finishAuthentication = mutation({
...
+    let verification;
+    try {
+      verification = await verifyAuthenticationResponse({
+        response,
+        expectedChallenge: toBase64URL(challengeRow.challenge),
+        expectedOrigin: args.expectedOrigin,
+        expectedRPID: args.expectedRpId,
+        credential: {
+          id: credentialId,
+          publicKey: new Uint8Array(passkey.publicKey),
```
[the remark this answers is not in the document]
Nicolapps: That's fine, we don't need backwards compatibility

packages/core/src/components/passkey/authentication.ts
```diff
@@ -235,7 +230,7 @@ export const finishAuthentication = mutation({
     const challengeRow = await consumeChallenge(
       ctx,
       "authentication",
-      clientData.challenge,
+      isoBase64URL.toBuffer(clientData.challenge),
```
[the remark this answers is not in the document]
Nicolapps: Already fixed in c9a36a5: `clientData.challenge` is now decoded **inside** the `okOrNull` block that parses the client data, so a missing or non-string `challenge` returns `PROTOCOL_ERROR` instead of throwing.

The `registration.ts` location got the same treatment.

(For the record, `isoBase64URL.toBuffer` does not actually throw on a *malformed* base64url string — `@hexagon/base64` decodes leniently and returns whatever bytes it can. It only throws when the value is not a string at all, which is the missing-`challenge` case this fix covers.)

packages/core/src/components/passkey/authentication.test.ts
```diff
@@ -131,6 +132,36 @@ describe("finishAuthentication", () => {
     expect(challenges).toEqual([]);
   });
 
+  test("accepts an assertion whose counter did not increase", async () => {
```
Nicolapps, line 135: I added a test here to codify that we do not check for counter growth, a security mechanism that can be used to detect duplicated passkeys.

SimpleWebauthn handles counter checks by default, but it is not very helpful to have it by itself. Applications that actually cares about duplicated passkeys must take an action like invalidating the passkey. We cannot do this safely from our providers what would be the desired outcome in the context of the app.

So I made our provider just ignore counter errors. If an application wanted to handle it better, its developers could create a custom provider on top of the passkey component.

packages/core/src/components/passkey/authentication.ts
```diff
@@ -179,7 +165,7 @@ export const finishAuthentication = mutation({
       );
       return { success: false, userError: { error: "PROTOCOL_ERROR" } };
     }
-    if (!authenticatorData.userPresent || !authenticatorData.userVerified) {
+    if (!authenticatorData.flags.up || !authenticatorData.flags.uv) {
       // The ceremony asks for `userVerification: "required"`, thus
       // `userVerified`/`userPresent` should be set
```
dowski: nit: update this comment since the names changed to `flags.up` and `flags.uv`.

packages/core/src/components/passkey/registration.ts
```diff
@@ -566,7 +585,7 @@ async function verifyAttestation(
     );
     return PROTOCOL_ERROR;
   }
-  if (!authenticatorData.userPresent || !authenticatorData.userVerified) {
+  if (!authenticatorData.flags.up || !authenticatorData.flags.uv) {
     // The ceremony asks for `userVerification: "required"`, thus
     // `userVerified`/`userPresent` should be set
```
dowski: Ditto on this comment.

packages/core/src/components/passkey/validation.ts
```diff
@@ -7,6 +7,16 @@ import { Infer, v } from "convex/values";
 // https://www.w3.org/TR/webauthn-3/#sctn-timeout-recommended-range
 export const CHALLENGE_TTL_MS = 10 * 60 * 1000; // 10 minutes
 
+// The COSE signature algorithms this provider accepts, as IANA identifiers
+// (https://www.iana.org/assignments/cose/cose.xhtml#algorithms). This list
+// is enforced when the attestation is verified. The client offers its own
+// `pubKeyCredParams` literal in `react.tsx`, so keep the two lists in sync
+// by hand. `@simplewebauthn/server` also verifies Ed25519 (-8).
```
dowski, line 14: Could they be kept somewhere common so they don't need to be kept in sync? Having to keep both sounds brittle.

dowski (approved):

packages/core/src/components/passkey/registration.ts
```diff
@@ -654,10 +665,9 @@ async function verifyAttestation(
   return {
     userError: null,
     credential: {
-      credentialId,
-      algorithm,
-      publicKey: toArrayBuffer(publicKey),
-      counter: authenticatorData.signatureCounter,
+      credentialId: storedCredentialId,
+      publicKey: toArrayBuffer(credential.publicKey),
```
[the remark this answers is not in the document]
Nicolapps: We don't need backwards compatibilty.

packages/core/src/components/passkey/validation.ts
Nicolapps: Yep, this state is only temporary. The list will be shared in the PR that uses SimpleWebAuthn on the client

packages/core/src/components/passkey/schema.ts
```diff
@@ -11,9 +11,7 @@ export default defineSchema({
     name: v.optional(v.string()),
     // The raw credential ID bytes (the WebAuthn `rawId`).
     credentialId: v.bytes(),
-    algorithm: v.union(v.literal("ES256"), v.literal("RS256")),
-    // The encoded public key for signature checks. The format is a SEC1
-    // uncompressed point for ES256, and PKCS#1 DER for RS256.
+    // The COSE public key (this contains the key, algorithm and curve).
```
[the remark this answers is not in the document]
Nicolapps: We don't need backwards compatibility yet as the framework is not GA yet. There is another breaking change in the same diff so it’s not possible for an app to be mistakenly upgraded.

packages/core/src/components/passkey/authentication.ts
Nicolapps: Done in 0a70758d3e241594296e4df060ee7c2fb27d30eb, thanks

packages/core/src/components/passkey/registration.ts
Nicolapps: Also done in 0a70758d3e241594296e4df060ee7c2fb27d30eb

packages/core/src/components/passkey/authentication.ts
```diff
@@ -270,46 +279,70 @@ export const finishAuthentication = mutation({
...
+    let verification;
+    try {
+      verification = await verifyAuthenticationResponse({
+        response,
+        expectedChallenge: toBase64URL(challengeRow.challenge),
+        expectedOrigin: args.expectedOrigin,
+        expectedRPID: args.expectedRpId,
+        credential: {
+          id: credentialId,
+          publicKey: new Uint8Array(passkey.publicKey),
```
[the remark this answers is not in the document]
Nicolapps: This is also fine, we don't need backwards compatibility in this change

PR #506, opened 2026-08-20
Report every oauth client failure through flowError

erquhart:
- Add a `rejected` flow error for sign-ins the app's own backend turns down, with the message from its `ConvexError`
- Set `flowError` on start and storage failures too, so a caller that ignores the `signIn` rejection still has something to show

### DIFF
--- a/packages/core/src/oauth/client.test.ts
+++ b/packages/core/src/oauth/client.test.ts
@@ -264,7 +264,7 @@ describe("OAuth client", () => {
     expect(client.getSnapshot().isAuthenticated).toBe(false);
   });
 
-  test("an app rejection with non-string data gets the default copy", async () => {
+  test("an app rejection with non-string data carries no message", async () => {
     window.history.replaceState(null, "", "/?convexAuthCode=code-1");
     const storage = new InMemoryStorage();
     seedPendingFlow(storage);
@@ -274,7 +274,9 @@ describe("OAuth client", () => {
     await client.init();
 
     await vi.waitFor(() => expect(flowError()?.code).toBe("rejected"));
-    expect(flowError()?.message).toBe("Sign-in was declined.");
+    // Only a string carries copy the app meant for the user, so there is
+    // nothing to show here.
+    expect(flowError()?.message).toBeUndefined();
   });
 
   test("a rejecting storage during redemption sets oauth_error", async () => {
--- a/packages/core/src/oauth/client.test.ts
+++ b/packages/core/src/oauth/client.test.ts
@@ -279,7 +279,7 @@ describe("OAuth client", () => {
     expect(flowError()?.message).toBeUndefined();
   });
 
-  test("a rejecting storage during redemption sets oauth_error", async () => {
+  test("a rejected storage read during redemption sets oauth_error", async () => {
     window.history.replaceState(null, "", "/?convexAuthCode=code-1");
     // Fail reads of the saved flow key, the way an async storage might. Reads
     // of the session tokens still work.
@@ -300,7 +300,7 @@ describe("OAuth client", () => {
     expect(client.getSnapshot().isLoading).toBe(false);
   });
 
-  test("a rejecting storage during error cleanup keeps the flow error", async () => {
+  test("a rejected storage removal during error cleanup keeps the flow error", async () => {
     window.history.replaceState(null, "", "/?convexAuthError=access_denied");
     const storage: TokenStorage = {
       getItem: () => null,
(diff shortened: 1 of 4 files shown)


packages/core/src/oauth/client.test.ts
```diff
@@ -243,6 +244,80 @@ describe("OAuth client", () => {
...
+
+    await client.init();
+
+    await vi.waitFor(() => expect(flowError()?.code).toBe("rejected"));
+    // Only a string is text the app meant for the user, so there is nothing
+    // to show here.
+    expect(flowError()?.message).toBeUndefined();
+  });
+
+  test("a rejecting storage during redemption sets oauth_error", async () => {
```
dowski: Should this say "a rejected storage read ..."?

packages/core/src/oauth/client.test.ts
```diff
@@ -243,6 +244,80 @@ describe("OAuth client", () => {
...
+    const { client, mutation, flowError } = setupOAuth({ storage });
+
+    await client.init();
+
+    await vi.waitFor(() => expect(flowError()?.code).toBe("oauth_error"));
+    expect(mutation).not.toHaveBeenCalled();
+    expect(client.getSnapshot().isLoading).toBe(false);
+  });
+
+  test("a rejecting storage during error cleanup keeps the flow error", async () => {
```
dowski: Similar: "a rejected storage removal ..."?

packages/core/src/oauth/client.ts
```diff
@@ -53,17 +54,25 @@ export type OauthProviderRefs = {
...
-  "access_denied" | "expired" | "oauth_error" | "invalid_flow";
+  "access_denied" | "expired" | "rejected" | "oauth_error" | "invalid_flow";
 
 /** Why a sign-in failed. */
 export type OauthFlowError = {
   /** Why the sign-in failed. */
   code: OauthFlowErrorCode;
+  /**
+   * Text to show the user. Only set when the app's backend threw a
+   * `ConvexError` whose `data` is a string. That string is this text.
```
dowski, line 73: The backend may or may not be able to localize the string. Fine for now, but maybe include a TODO for tracking i18n long tail items.

packages/core/src/oauth/client.ts
```diff
@@ -190,6 +199,19 @@ async function takePendingFlow(
   }
 }
 
+/**
+ * Remove the saved sign-in flow after it has ended. Callers don't await this,
+ * so a rejecting storage is ignored instead of becoming an unhandled
```
dowski: Here's this "a rejecting storage" term again - maybe it's a real term that I just don't understand.

packages/core/src/oauth/client.ts
```diff
@@ -203,27 +225,54 @@ export function oauth(): AmbientSignInClient {
...
+        // becomes a flow error like any other failure here.
         try {
+          const pending = await takePendingFlow(storage);
+          if (pending === null) {
+            setFlowError("invalid_flow");
+            return false;
+          }
+          const completeSignIn = makeFunctionReference<"mutation">(
+            pending.completeSignIn,
+          ) as OauthProviderApi["completeSignIn"];
```
dowski, line 279: I think this is that thing I was asking about on an earlier PR re: the string name of the function and if we could avoid keeping it. It feels weird to do that and then hydrate it back to a function ref here in client code.

Can we have a TODO to see about cleaning this up?

packages/core/src/oauth/client.test.ts
dowski: Also I wonder, what would cause storage operations to get rejected?

dowski (approved):

packages/core/src/oauth/client.ts
erquhart: "rejecting" as in a rejected promise (async `storage.remove()`)

packages/core/src/oauth/client.test.ts
erquhart: "Rejecting" as in the overall process that begins with promise rejection and ends with a failed flow and error message. Can change it.

The storage method could be a number of things, especially for RN. This is just making sure there's a path if a storage method throws unexpectedly.

packages/core/src/oauth/client.ts
erquhart: just realized in your earlier comment you mean the verb tense. I can change it.

packages/core/src/oauth/client.ts
erquhart: This was actually an intentional decision - without this you have to configure the oauth() function with each provider's set of function refs on the client, and I really wanted to avoid adding weight to the client config. I get the "feels weird" part, but it should be solid. Will rework if preferred.

erquhart: added todo

PR #508, opened 2026-08-20
explore using a function type for contravariance

ianmacartney:
----


ianmacartney:
@dowski this approach has some awkwardness to needing to use `userCallback(` but might be interesting. 

I wonder if the circular imports go away entirely when it's done as component configuration, since there won't be a function getting re-exported. Just passing in the user callback to the component directly and calling the component from the client directly?

ianmacartney:
Other approach in #509 I think is better?

PR #511, opened 2026-08-21
Add react-oauth example app

erquhart:
- Add basic oauth example app
- Uses GitHub and Google
- Allows account linking with an explanation that its a workaround in lieu of official support
- Will follow up with tests for the example


examples/react-oauth/convex/auth.ts
```diff
@@ -0,0 +1,17 @@
...
+
+const core = setupCore({ component: components.auth });
+export const { signOut, refreshSession, isAuthenticated } = core;
+
+export const { startSignInGoogle, completeSignInGoogle } = setupGoogle(core, {
+  component: components.oauthGoogle,
+  allowedRedirectOrigins: ["http://localhost:5173"],
+}).attachUserCallbacks({ createUser: internal.users.createUserGoogle });
+
+export const { startSignInGithub, completeSignInGithub } = setupGithub(core, {
```
Nicolapps, line 14: nit: shouldn’t this be `setupGitHub`?

examples/react-oauth/vite.config.ts
```diff
@@ -0,0 +1,6 @@
+import react from "@vitejs/plugin-react";
+import { defineConfig } from "vite";
+
+export default defineConfig({
+  plugins: [react()],
```
Nicolapps, line 5: Since the demo code hardcodes `5173` in the redirect URLs, we should make the port a constant here. Could you please copy the code from the passkey demo’s `vite.config.ts` file?

examples/react-oauth/convex/users.ts
```diff
@@ -0,0 +1,75 @@
...
+    return await createUser(ctx, {
+      email: args.profile.email,
+      emailVerified: args.profile.emailVerified,
+    });
+  },
+});
+
+async function createUser(
+  ctx: MutationCtx,
+  args: { email?: string; emailVerified: boolean },
```
Nicolapps, line 55: The current API of the OAuth component looks footgunny (users could forget to check the emailVerified field). What do you think about having something like `{ emailVerified?: string } | { emailUnverified?: string }`?

examples/react-oauth/convex/schema.ts
```diff
@@ -0,0 +1,8 @@
+import { defineSchema, defineTable } from "convex/server";
+import { v } from "convex/values";
+
+export default defineSchema({
+  users: defineTable({
+    email: v.string(),
```
Nicolapps: What do you think about naming the field `verifiedEmail` so that it incentivizes users to never store unvalidated email addresses here?

examples/react-oauth/convex/users.ts
```diff
@@ -0,0 +1,75 @@
...
+) {
+  const { email, emailVerified } = args;
+  if (email === undefined) {
+    throw new ConvexError("No email provided");
+  }
+  // Linking accounts is technically not supported, but you can do it manually
+  // if you choose. For this example we consider GitHub and Google's email
+  // verification status to be sufficient to link two accounts based on email.
```
Nicolapps, line 63: What do you mean by “Linking accounts is technically not supported”?

Maybe something like this would work:

```suggestion
  // Manually linking accounts that are using the same provider-verified email address
```

Nicolapps (approved):

examples/react-oauth/convex/auth.ts
erquhart: The pattern I've been following is that caps are for word separations, same reason we do  oauthGoogle instead of oAuthGoogle or OAuthGoogle

examples/react-oauth/convex/users.ts
erquhart: I don't actually think we should encourage this approach at all, its just a stop gap for proper account linking. If we wanted to make it safer in the interim I'd just not pass through any unverified emails - but for an alpha that documents "make sure you do this" I feel like its the right amount of caution. Let me know what you think.

examples/react-oauth/convex/users.ts
erquhart: I'm saying its not supported because the user needs to understand that they should know what they're doing if they follow this approach (eg., check for emailVerified)- we haven't actually thought through all the guardrails that may be needed. That's also why I don't want to start adding small guardrails here, its kinda rough because it isn't actually a supported behavior yet, just happens to work.

I'd rather aim at a lower level, eg., unverified emails shouldn't even make it into a user function without them setting some sort of `dangerously` config. But I also don't want to get ahead and try to do that today without giving it proper consideration in context of account linking.

PR #517, opened 2026-08-24
Add docs for OAuth providers

dowski:
Also includes some tweaks to the general docs that better reflect the current state.

### DIFF
--- a/packages/docs/content/docs/getting-started.mdx
+++ b/packages/docs/content/docs/getting-started.mdx
@@ -45,7 +45,7 @@ Add the Convex Auth package to your project:
 {/* prettier-ignore-end */}
 
 <Callout type="warn">
-  Convex Auth v2 is still in development. This installs the latest `reboot`
+  Convex Auth v2 is still in development. This installs the latest development
   build from a preview registry rather than a published release.
 </Callout>
 
--- a/packages/docs/content/docs/login-providers/oauth.mdx
+++ b/packages/docs/content/docs/login-providers/oauth.mdx
@@ -1,0 +1,327 @@
+---
+title: OAuth (Google and GitHub)
+description: Allow users to sign-in with an account on a supported provider.
+---
+
+import { Step, Steps } from "fumadocs-ui/components/steps";
+import { Tab, Tabs } from "fumadocs-ui/components/tabs";
+
+The `oauth` component and supporting provider functions (`setupGoogle`,
+`setupGithub`) allow your users to login via their existing Google or GitHub
+accounts.
+
+The steps below will guide you through configuring Google and/or GitHub
+login. In the end, your Convex deployment will host HTTP routes for handling
+the OAuth callbacks and redirect back to your application for handling the result.
+
+<Callout type="info">
+  Support for using other OAuth identity providers isn't included yet.
+</Callout>
+
+## Common setup
+
+These steps are the same whichever provider you use. Install the `oauth`
+component once per identity provider, so an app offering both Google and GitHub
+login repeats the provider-specific steps below for each of them.
+
+<Steps>
+
+<Step>
+
+### Follow the “Getting Started” guide
+
+First, follow the [Getting Started](/getting-started) guide:
+
+<Cards>
+  <Card
+    href="/getting-started"
+    title="Getting Started"
+    description="Installing dependencies + initial set up."
+  />
+</Cards>
+
+</Step>
+
+<Step>
+
+### Add a `users` table to the schema
+
+Add an empty `users` table to your schema in `convex/schema.ts`.
+
+Convex Auth will create a row for each account. You can also add application-specific fields to this table if necessary.
+
+The table does not have to be called `users` — if yours has another name, pass
+it to `setupCore`, and every provider's `attachUserCallbacks` will then require
+callbacks typed with that table's ids:
+
+```ts title="convex/auth.ts"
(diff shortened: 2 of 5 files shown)


dowski:
* **#517** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/517?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/517?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#518** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/518?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `reboot`


<h2></h2>

packages/docs/content/docs/getting-started.mdx
```diff
@@ -22,31 +22,31 @@ Add the Convex Auth package to your project:
...
     ```
   </Tab>
 </Tabs>
 
 {/* prettier-ignore-end */}
 
 <Callout type="warn">
-  Convex Auth v2 is still in development. This installs the latest `reboot`
-  build from a preview registry rather than a published release.
+  Convex Auth v2 is still in development. This installs a preview version from
```
Nicolapps: See https://convexdev.slack.com/archives/C04U4LQ9WDD/p1787584389711279?thread_ts=1787576500.633369&cid=C04U4LQ9WDD

packages/docs/content/docs/login-providers/oauth.mdx
```diff
@@ -0,0 +1,179 @@
...
+For each provider that you configured, write a `createUser[Google|Github]`
+mutation that creates a row in your `users` table and returns its id. The
+provider calls it the first time it sees an account.
+
+In the example below, both provider mutations call into a common `createUser`
+function which allows sharing the creation logic.
+
+<include
+  lang="ts"
+  meta='title="convex/users.ts" highlight="createUserGoogle = internalMutation({...});" highlight="createUserGithub = internalMutation({...});"'
```
Nicolapps: What do you think about renaming the current `react-oauth` example to `react-oauth-multiple` (or equivalent) and creating a simpler demo that only supports a single OAuth method? The current code examples are more complicated than necessary. Also it shows how to do manual account linking, which we want to avoid for now

(If we also want to document how to install multiple providers, we could have a dedicated docs page for it)

examples/react-oauth/src/App.tsx
```diff
@@ -20,6 +20,7 @@ const FLOW_ERROR_COPY: Record<OauthFlowErrorCode, string> = {
   invalid_flow: "This sign-in can't be completed here. Please try again.",
 };
 
+//#region login
```
Nicolapps: `FLOW_ERROR_COPY` should definitely be included in the code sample that folks are going to copy/paste from. Having imports would be useful too.

Can we please include the entire file? If it looks too complicated, we should consider having a React Router-based demo and only sharing the code for the sign in page.

Also, please feel free to remove GoogleIcon/GitHubIcon from the icons to simplify the user-visible file

examples/react-oauth/convex/users.ts
```diff
@@ -50,6 +31,25 @@ export const createUserGithub = internalMutation({
...
+  args: {},
+  handler: async (ctx) => {
+    const identity = await ctx.auth.getUserIdentity();
+    if (identity === null) {
+      return null;
+    }
+    const userId = ctx.db.normalizeId("users", identity.subject);
+    if (userId === null) {
+      return null;
+    }
```
Nicolapps: (This should be using getAuthUserId when it's merged: https://github.com/get-convex/convex-auth/pull/515)

packages/docs/content/docs/login-providers/oauth.mdx
```diff
@@ -0,0 +1,179 @@
+---
+title: OAuth (Google and GitHub)
+description: Allow users to sign-in with an account on a supported provider.
+---
+
+import { Step, Steps } from "fumadocs-ui/components/steps";
+
+The `oauth` component and supporting provider functions (`setupGoogle`,
+`setupGithub`) allow your users to login via their existing Google or GitHub
+accounts.
```
Nicolapps, line 11: This page doesn't explain clearly how it works if developers want to set up non-Google/GitHub OAuth flows.

If it's something that's not supported yet it's fine, but we should mention that so that it's not confusing to beta users

examples/react-oauth/src/App.tsx
dowski: Done.

packages/docs/content/docs/getting-started.mdx
dowski: I'll just revert this. We can share the alpha release link w/ the hackathon and leave this in the docs for now for anyone who wants the bleeding edge.

packages/docs/content/docs/login-providers/oauth.mdx
dowski: Thanks, I split up the examples.

packages/docs/content/docs/login-providers/oauth.mdx
dowski: Done.

dowski:
### Merge activity

* **Aug 24, 6:56 PM UTC**: @dowski merged this pull request with [Graphite](https://app.graphite.com/github/pr/get-convex/convex-auth/517).

PR #521, opened 2026-08-24
add release scripts

erquhart:
Adds alpha and latest release scripts, core only. Follows existing pattern from convex auth v1 and other components.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,107 @@
+# Developing guide
+
+## Running locally
+
+```sh
+pnpm install
+```
+
+Then run one of the example apps from its own directory:
+
+```sh
+cd examples/react-password
+pnpm dev
+```
+
+Each example is a self-contained Convex backend that mounts the auth components,
+so the first `pnpm dev` prompts you to pick or create a deployment.
+
+- `examples/react-minimal` uses the core component with the anonymous provider.
+- `examples/react-password` adds the password provider.
+- `examples/react-passkey` adds the passkey provider.
+- `examples/react-github` adds the GitHub Oauth provider.
+- `examples/react-google` adds the Google OAuth provider.
+- `examples/nextjs` covers the Next.js integration.
+
+## Common tasks
+
+- `pnpm build` — compile `@convex-dev/auth` to `dist/` (also runs on install
+  via `prepare`).
+- `pnpm test` — run tests across packages (Vitest).
+- `pnpm typecheck` — typecheck across packages.
+- `pnpm lint` — ESLint.
+- `pnpm format` — Prettier.
+
+## Testing
+
+```sh
+pnpm build
+pnpm lint
+pnpm typecheck
+pnpm test
+```
+
+`pnpm typecheck` covers `packages/argon2id-wasm`, which needs the generated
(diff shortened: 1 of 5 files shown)


dowski (commented):
Would you add a Publishing section to the README that covers how to use these commands?

CONTRIBUTING.md
```diff
@@ -0,0 +1,106 @@
...
+pnpm dev
+```
+
+Each example is a self-contained Convex backend that mounts the auth components,
+so the first `pnpm dev` prompts you to pick or create a deployment.
+
+- `examples/react-minimal` uses the core component with the anonymous provider.
+- `examples/react-password` adds the password provider.
+- `examples/react-passkey` adds the passkey provider.
+- `examples/react-oauth` adds the oauth provider.
```
dowski: Note this is now split into `react-github` and `react-google`.

packages/core/package.json
```diff
@@ -122,11 +118,15 @@
...
+    "alpha": "pnpm version prerelease --preid alpha && pnpm publish --tag alpha --publish-branch reboot && git push --follow-tags",
     "build": "tsc -p tsconfig.build.json",
     "clean": "rm -rf dist",
     "generate:common-passwords": "bun src/components/password/scripts/generateCommonPasswords.script.ts",
     "prepare": "pnpm run build",
-    "typecheck": "tsc --noEmit"
+    "preversion": "pnpm -w install --frozen-lockfile && pnpm -w build && pnpm -w lint && pnpm -r --filter=!argon2id-wasm typecheck && pnpm -w test",
+    "release": "pnpm version patch && pnpm publish --publish-branch reboot && git push --follow-tags",
+    "typecheck": "tsc --noEmit",
+    "version": "(npm whoami || npm login) && vim -c 'normal o' -c 'normal o## '$npm_package_version ../../CHANGELOG.md && prettier -w ../../CHANGELOG.md && git add ../../CHANGELOG.md"
```
dowski: So this is used internally by `alpha` and `release`? What does it do? Maybe add a comment as it's not fully clear to me.

CONTRIBUTING.md
```diff
@@ -0,0 +1,106 @@
...
+
+```sh
+pnpm alpha
+```
+
+For a standard (latest tag) release:
+
+```sh
+pnpm release
+```
```
dowski, line 93: Is there any `--dry-run` option or anything?

dowski (approved):
Thanks Shawn!

packages/core/package.json
erquhart: Yeah, its the same one we use in all of the component packages, it makes sure you're logged in to npm so you don't get halfway through the release process and error out, and it opens vim with a new section ready for changelog entry. Json doesn't support comments, but this is standard across our packages.

CONTRIBUTING.md
erquhart: Not currently, guessing we don't want to hold on it for now, this is just how we've done these for component packages. I can take a task to add dry running if we want that.

CONTRIBUTING.md
erquhart: will update

PR #522, opened 2026-08-24
Add some refresh token reuse detection

dowski:
This prevents a stolen token from being used to refresh a session and lock out a legitimate client.

There's more hardening we can do here but this at least closes the main avenue where an attacker could maintain a session with a stolen token. A concurrently running regular client refresh would eventually trip the reuse detection and delete the session.

This involves tracking previously spent refresh token hashes in a new table, and no longer storing an individual previously used refresh hash in the sessions table.

### DIFF
--- a/packages/core/src/components/core/public.ts
+++ b/packages/core/src/components/core/public.ts
@@ -621,6 +621,10 @@ export const refresh = mutation({
     if (!isValid) {
       // Past its refresh-token lifetime or a spent token past the grace
       // window: delete the session and grant no more tokens.
+      console.warn(
+        "Spent refresh token used out of grace window: deleting the associated session. " +
+          "This could be due to a bug or a leaked refresh token.",
+      );
       await deleteSession(ctx, session._id);
       return null;
     }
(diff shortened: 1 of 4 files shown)


dowski:
* **#555** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/555?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#522** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/522?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/522?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `reboot`


<h2></h2>

dowski:
Going to touch this up a bit more before reviews.

packages/core/src/components/core/public.ts
```diff
@@ -424,62 +606,27 @@ export const refresh = mutation({
...
-    if (!session) return null;
-    if (session.refreshTokenExpiresAt < now) {
-      // Past its refresh-token lifetime: delete the dead row and report no
-      // session.
-      await ctx.db.delete("sessions", session._id);
+    if (session === null) return null;
+    if (!isValid) {
+      // Past its refresh-token lifetime or a spent token past the grace
+      // window: delete the session and grant no more tokens.
+      await deleteSession(ctx, session._id);
```
Nicolapps, line 628: Do you think it might be worth logging with `console.warn` here when we do so? It might help us diagnose issues if it turns out that we’re deleting sessions unexpectedly

Nicolapps (approved):

packages/core/src/components/core/public.ts
dowski: Good call - done.

PR #558, opened 2026-08-26
Use FunctionReference_future for user callbacks

dowski:
You can see how it benefits in two ways in this change:

1. The nextjs example now shares createUser and onSignIn callbacks
2. The react-passkey and react-passkey-basic examples had a bug that the covariant FunctionReference previously allowed

This required updating to `convex 1.46.0`.


dowski:
* **#558** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/558?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/558?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `reboot`


<h2></h2>

packages/core/src/lib/types.ts
```diff
@@ -200,27 +203,26 @@ export const vOnSignIn = v.object({
...
-> = FunctionReference<
+> = ContravariantFunctionReference<
   "mutation",
   "internal",
-  {
-    provider: Provider;
-    providerAccountId: string;
-    profile: Profile;
-  },
+  { provider: Provider; providerAccountId: string; profile: Profile },
```
ianmacartney, line 210: I suspect if you want to have discriminated types for each provider, you will need to put the provider and profile on the same nested object field.   
Today you can't do this as your user fn args:  
`args: v.union(v.object({provider: v.literal("google"), profile: vGoogleFields}), v.object({provider: v.literal("emailPassword"), profile: vEmailFields}))` because args can't be a union at the top level. So you'd be forced to do:

```typescript
args: {
  provider: v.union(v.literal("google"), v.literal("emailPassword")),
  profile: v.union(vGoogleFields, vEmailFields),
}
```

which then doesn't have a way to easily handle each set of fields.  
Instead I think we should do something like

```typescript
args: {
  providerMetadata: v.union(
    v.object({ provider: v.literal("google"), profile: vGoogleFields}),
    v.object({ provider: v.literal("emailPassword"), profile: vEmailFields}),
  ),
  ...
}
```

or some other naming scheme.

Or we start allowing top-level unions of arguments

dowski: Thanks for this - I landed the fix in https://github.com/get-convex/convex-auth/pull/620.

packages/core/package.json
```diff
@@ -202,7 +202,7 @@
     "@types/babel__code-frame": "^7.27.0",
     "@types/node": "^24.12.4",
     "@types/react": "^19.2.17",
-    "convex": "^1.43.0",
+    "convex": "^1.46.0",
```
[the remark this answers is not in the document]
dowski: Resovled.

examples/react-passkey/convex/users.ts
```diff
@@ -10,7 +10,7 @@ export const createUser = internalMutation({
     provider: v.object({
       name: v.literal("passkey"),
       accountId: v.string(),
-      profile: v.object({ username: v.string() }),
+      profile: v.object({ username: v.union(v.string(), v.null()) }),
```
Nicolapps, line 13: Why is this change necessary? Isn’t it guaranteed that the username is set at this point?

Nicolapps (approved):

examples/react-passkey/convex/users.ts
dowski: The result here can be `null` so it flows through the type. Maybe we want to throw if that's the case at that point in the code? It seems like it would be a programming/framework error if there was no username at that point. WDYT?

dowski: Here's the code I'm talking about:  
  
https://github.com/get-convex/convex-auth/blob/71e6b234d18413166eecfff947e7f9bca33b24bc/packages/core/src/components/passkey/setup.ts#L470

dowski: Resolved out-of-band. We'll actually be removing this data from the profile object (`username` should be queried from the component when needed).

PR #567, opened 2026-08-27
[no-op] authPasskey: name the finish functions in the comments

Nicolapps:

PR #556, opened 2026-08-26
usernamePasskey: limit the number of passkeys of one user

Nicolapps:
This modifies the passkey provider functions to add a limit of how many passkeys a given user can have. The goal is to avoid a user having too many passkeys registered, which might be an indicator of abuse and hit Convex query limits. This is enforced at the app level, not at the component level.

PR #570, opened 2026-08-27
passkey: reorganize usePasskey for future extensibility

Nicolapps:
Currently, apps that use the passkey component use a single React hook, `usePasskey`, which manages the entire login page. This means choosing whether to register a new passkey or request an existing one, plus autofill support.

This structure isn’t very extensible. We want to separate the standard flow behavior from the autofill behavior. This will make it easier to implement:
- the upcoming email flow (using email instead of username)
- passkey management (e.g. a logged in user adding or deleting a passkey)
- user-authored custom flows (e.g. if a library user want to implement a “passkey-as-2FA” flow)

This PR refactors the `usePasskey` hook the following way:

- `usePasskey`: manages the entire login page flow, soon to be renamed to `useUsernamePasskeySignIn`
  - `usePasskeyAutofill` (autofill behavior)
  - `usePasskeyCeremonySlot` (helper for coordinating an auth flow with the autofill behavior)
    - `runSignInOrSignUpFlow` (the implementation of the standard flow)


Nicolapps:
* **#580** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/580?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#570** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/570?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/570?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#505** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/505?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `reboot`


<h2></h2>

examples/react-passkey-basic/src/routes/logIn.tsx
```diff
@@ -76,8 +76,12 @@ function errorMessage(
   userError:
     | Extract<PasskeySignInResult, { success: false }>["userError"]
     | PasskeyAutofillError,
-): string {
+): string | null {
   switch (userError.error) {
+    case "ALREADY_PENDING":
```
Nicolapps, line 81: I’m not fully convinced about this change: I added a TODO(nicolas) comment to possibly change the structure so that we don't return any value when encountering this case.

packages/core/src/components/passkey/react.tsx
```diff
@@ -245,434 +83,85 @@ function autofillAbortReason(
...
+      /**
+       * The state of the autofill flow:
+       *
+       * - `available`: whether the browser can offer passkeys in the
+       *   autocompletion list; `null` while the detection runs.
+       * - `status`: the state of the request; see
+       *   {@link PasskeyAutofillStatus}.
+       * - `lastError`: the most recent failure, for display or logging.
+       */
+      autofill: {
```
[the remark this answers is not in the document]
Nicolapps: We don't need backward compatibility

packages/core/src/components/passkey/react.tsx
```diff
@@ -245,434 +83,85 @@ function autofillAbortReason(
...
-      // Never leave a `pause()` caller hanging, whatever path exited the
-      // loop.
-      onParked?.();
-      onParked = null;
-      // Retract the handle, unless a newer effect run published its own.
-      if (autofillHandleRef.current === handle) {
-        autofillHandleRef.current = null;
-      }
-    });
+      await setSession(result.tokens);
```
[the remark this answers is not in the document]
Nicolapps: That's fine, both sign in attempts are valid so either order is fine

packages/core/src/components/passkey/flows.ts
```diff
@@ -0,0 +1,200 @@
+/**
+ * The passkey sign-in flow: its mutation and ceremony sequence, with no
+ * React. The hooks of `react.tsx` run it through their ceremony slot.
+ * The autofill flow has no entry here: its ceremony
+ * runs in the request loop of `usePasskeyAutofill`, which calls the two
+ * mutations around it.
```
dowski: I'm not sure if this is a Claude comment, but I find that it likes to document things that aren't present. I find that confusing. When I'm reading a doc comment I want to know about what's present, not what's not present.

packages/core/src/components/passkey/flows.ts
```diff
@@ -0,0 +1,200 @@
...
+} from "./client.ts";
+import type {
+  FinishSignInResult,
+  FinishSignUpResult,
+  StartAutofillSignInResult,
+  StartSignInResult,
+} from "./setup.ts";
+
+/**
+ * The `startSignIn` mutation the app re-exports from its `setupCore`.
```
dowski: These aren't exported by `setupCore` (here and in the few refs below).

packages/core/src/components/passkey/flows.ts
```diff
@@ -0,0 +1,200 @@
...
+ * mutations around it.
+ *
+ * Internal: the package blocks the
+ * `@convex-dev/auth/providers/passkey/flows` path. `react.tsx` re-exports
+ * the types an app needs.
+ *
+ * @module
+ */
+
+import type { ConvexReactClient } from "convex/react";
```
dowski, line 8: The doc comment above says "no React" but then we import `ConvexReactClient`. Perhaps this could actually be React-agnostic or the doc comment should be updated.

packages/core/src/components/passkey/react_impl.tsx
```diff
@@ -0,0 +1,470 @@
...
+ * by the autofill loop.
+ */
+export function usePasskeyCeremonySlot(options: {
+  autofill: PasskeyAutofillGates;
+}) {
+  // The two gates, not the object that carries them: the whole return
+  // value of `usePasskeyAutofill` changes identity whenever the reported
+  // state does, while each gate keeps its identity for the life of that
+  // hook. Reading them here is what lets `run` stay stable.
+  const { pause, resume } = options.autofill;
```
dowski: Could these be named `pauseAutofillFlow` and `resumeAutofillFlow` or something? Reading them where they are called later, it's not as clear what it is that they're pausing and resuming.

packages/core/src/components/passkey/react_impl.tsx
```diff
@@ -0,0 +1,470 @@
...
+  }, []);
+
+  const resume = useCallback(() => {
+    pauseCountRef.current = Math.max(0, pauseCountRef.current - 1);
+    if (pauseCountRef.current === 0) {
+      wakeRef.current?.();
+    }
+  }, []);
+
+  useEffect(() => {
```
dowski, line 181: I _think_ I understand what's going on in here. :laughing:

However I wonder: does all of this have to run in an effect? Could the autofilll be triggered by a focus event in the associated `<input>`? I guess maybe the effect is required no matter what since we're bridging React with browser APIs.

dowski (approved):

packages/core/src/components/passkey/react_impl.tsx
Nicolapps: Renamed to `pauseAutofillFlow`/`resumeAutofillFlow` in `PasskeyAutofillGates` and everywhere they're used/tested.

Verified the upstack PRs still typecheck with this rename (cherry-picked the commit onto each). One heads-up: `nicolas/passkey-management-hooks` will hit a small merge conflict on `options.autofill?.pause?.()` / `resume?.()` when it gets restacked onto this — trivial rename to the same optional-call form, just flagging it since that branch is a live worktree I didn't want to touch.

packages/core/src/components/passkey/flows.ts
Nicolapps: Thanks, I removed these sloppy comments

packages/core/src/components/passkey/flows.ts
Nicolapps: I strongly agree. I think the model is overfitting on its instructions. I try to catch these during my reviews but this one slipped through. It’s now fixed

packages/core/src/components/passkey/flows.ts
Nicolapps: Also removed

packages/core/src/components/passkey/react_impl.tsx
Nicolapps: (ugh why did it reply on its own)

packages/core/src/components/passkey/react_impl.tsx
Nicolapps: From everything I read, starting the ceremony when the page loads is what is expected. See for instance https://passkeys.dev/docs/use-cases/bootstrapping/

I think it's done that way so that there is no latency when the user selects the input

Nicolapps: The docs of SimpleWebAuthn also give [clearer guidance](https://simplewebauthn.dev/docs/packages/browser#browser-autofill-conditional-ui) about this:

> **Guidance from platform vendors indicates that it is important to initialize WebAuthn's Conditional UI experience as soon as possible.** Placing this logic in <head> should give browsers enough time to query authenticators for any discoverable credentials to display to the user. It may also work to delay painting UI For N milliseconds to achieve the same thing in something like a single-page app that would trigger this experience some time after page load. It's still early days for this capability, though, and different browsers may have different quirks while the feature evolves.
> 
> This new "pending" WebAuthn request will be automatically cancelled on a subsequent execution of startAuthentication() in, for example, a click handler that triggers the browser's typical "modal" WebAuthn experience.
>
> Conditional UI is still a nascent capability, but this method should be pretty reliable since the API is largely settled. This new logic has been successfully tested in both Chrome Canary v103 and Safari 16.0 in the macOS Ventura beta.

PR #579, opened 2026-08-28
Move the test authenticator to a private package

Nicolapps:
This PR refactors the internal tests so that the test authenticator is no longer exported by the `@convex-dev/auth` package.

PR #577, opened 2026-08-27
authPasskey: split the registration functions per flow

Nicolapps:
This PR implements Christian’s proposal in https://github.com/get-convex/convex-auth/pull/460#discussion_r3804459383. 

<table>
<thead>
<tr>
<th>Before</th>
<th>After</th>
</tr>
</thead>
<tbody>
<tr>
<td rowspan="2"><code>startRegistration({ userId: string | null })</code></td>
<td><code>startRegistrationForNewUser({})</code></td>
</tr>
<tr>
<td><code>startRegistrationForExistingUser({ verifiedUserId })</code></td>
</tr>
<tr>
<td rowspan="2"><code>finishRegistration({ verifiedUserId })</code></td>
<td><code>finishRegistrationForNewUser({ newUserId })</code></td>
</tr>
<tr>
<td><code>finishRegistrationForExistingUser({ verifiedUserId })</code></td>
</tr>
</tbody>
</table>

This allows the calling code to be more explicit and avoid issues with flow confusion (we now return `PROTOCOL_ERROR` if finishing a different kind of flow than the one that was started).

We also now return a `PROTOCOL_ERROR` if the existing user flow completes with a different user ID than the one that started it. Previously the code assumed it would imply a developer error, but I realized that it can happen if the user logs out and then logs in with another account mid-flow.

PR #571, opened 2026-08-27
passkey: rename usePasskey to useUsernamePasskeySignIn

Nicolapps:
This PR is a no-op refactor that renames the `usePasskey` hook. The name of the hook was too generic, while this hook specifically handles the identify-first login page for “username + passkey.”

PR #580, opened 2026-08-30
passkey: move the wire to WebAuthn JSON

Nicolapps:
This changes the format of the argument/return values used by the passkey component and provider. Instead of using `v.bytes()`, we now use objects similar to [`RegistrationResponseJSON`](https://w3c.github.io/webauthn/#dictdef-registrationresponsejson).

This will make it easier to switch to using SimpleWebAuthn everywhere, because it will allow us to use these request/response structures directly instead of converting to/from our custom wire format.

Nothing in this wire format is proprietary to SimpleWebAuthn, so if we ever wanted to change the underlying implementation in the future we wouldn’t need any breaking change.

### DIFF
--- a/packages/core/src/components/passkey/authentication.ts
+++ b/packages/core/src/components/passkey/authentication.ts
@@ -1,8 +1,6 @@
-import type {
-  AuthenticationResponseJSON,
-  AuthenticatorTransportFuture,
-} from "@simplewebauthn/server";
+import type { AuthenticatorTransportFuture } from "@simplewebauthn/server";
 import { verifyAuthenticationResponse } from "@simplewebauthn/server";
+import { vAuthenticationResponseJSON } from "./validation.ts";
 import {
   decodeClientDataJSON,
   isoBase64URL,
@@ -16,6 +14,7 @@ import {
   okOrNull,
   randomChallenge,
   rpIdHashMatches,
+  toArrayBuffer,
 } from "./helpers.ts";
 import {
   credentialDescriptor,
@@ -126,26 +125,28 @@ export const finishAuthentication = mutation({
     purpose: v.string(),
     expectedRpId: v.string(),
     expectedOrigin: v.string(),
-    credentialId: v.bytes(),
-    authenticatorData: v.bytes(),
-    clientDataJSON: v.bytes(),
-    signature: v.bytes(),
+    response: vAuthenticationResponseJSON,
   },
   returns: finishAuthenticationResult,
   handler: async (ctx, args): Promise<FinishAuthenticationResult> => {
     validatePurpose(args.purpose);
+    // `verifyAuthenticationResponse` checks that `id` and `rawId` agree, so
+    // either one can drive the lookup.
+    const credentialId = toArrayBuffer(
+      isoBase64URL.toBuffer(args.response.rawId),
+    );
     const passkey = await ctx.db
       .query("passkeys")
-      .withIndex("by_credentialId", (q) =>
-        q.eq("credentialId", args.credentialId),
-      )
+      .withIndex("by_credentialId", (q) => q.eq("credentialId", credentialId))
       .first();
     if (passkey === null) {
       return { success: false, userError: { error: "UNKNOWN_CREDENTIAL" } };
     }
 
     const authenticatorData = okOrNull(() =>
-      parseAuthenticatorData(new Uint8Array(args.authenticatorData)),
+      parseAuthenticatorData(
+        isoBase64URL.toBuffer(args.response.response.authenticatorData),
+      ),
     );
     if (authenticatorData === null) {
       console.warn(
@@ -183,7 +184,7 @@ export const finishAuthentication = mutation({
     //               callback variant of the `expectedChallenge`
     //               argument of `verifyAuthenticationResponse`.`
     const clientData = okOrNull(() =>
-      decodeClientDataJSON(toBase64URL(args.clientDataJSON)),
+      decodeClientDataJSON(args.response.response.clientDataJSON),
     );
 
     // Here we perform a few checks manually. These checks are also done later by
@@ -279,28 +280,15 @@ export const finishAuthentication = mutation({
       return { success: false, userError: { error: "PROTOCOL_ERROR" } };
     }
 
-    const credentialId = toBase64URL(args.credentialId);
-    const response: AuthenticationResponseJSON = {
-      id: credentialId,
-      rawId: credentialId,
-      response: {
-        clientDataJSON: toBase64URL(args.clientDataJSON),
-        authenticatorData: toBase64URL(args.authenticatorData),
-        signature: toBase64URL(args.signature),
-      },
-      clientExtensionResults: {},
-      type: "public-key",
-    };
-
     let verification;
     try {
       verification = await verifyAuthenticationResponse({
-        response,
+        response: args.response,
         expectedChallenge: toBase64URL(challengeRow.challenge),
         expectedOrigin: args.expectedOrigin,
         expectedRPID: args.expectedRpId,
         credential: {
-          id: credentialId,
+          id: args.response.rawId,
           publicKey: new Uint8Array(passkey.publicKey),
           // Passkey attestations can carry a counter that can be used by
           // applications to detect duplicated passkeys. The goal is to let
--- a/packages/core/src/components/passkey/authentication.ts
+++ b/packages/core/src/components/passkey/authentication.ts
@@ -130,9 +130,16 @@ export const finishAuthentication = mutation({
   returns: finishAuthenticationResult,
   handler: async (ctx, args): Promise<FinishAuthenticationResult> => {
     validatePurpose(args.purpose);
-    // `verifyAuthenticationResponse` checks that `id` and `rawId` agree, so
-    // either one can drive the lookup.
     const credentialId = toArrayBuffer(
+      // args.response contains the credential ID twice: in `args.response.id`
+      // and `args.response.rawId`. They are identical by definition (and SimpleWebAuthn
+      // checks for their equality), so we can use either one here.
+      // Why do we send the same data twice over our wire format? Because our
+      // wire format matches the `AuthenticationResponseJSON` DOM type
+      // (https://w3c.github.io/webauthn/#dictdef-authenticationresponsejson).
+      // Using the DOM type as the wire format makes the implementation easier both on
+      // the client (argument serialization is simple) and on the server
+      // (SimpleWebAuthn accepts `args.response` directly).
       isoBase64URL.toBuffer(args.response.rawId),
     );
     const passkey = await ctx.db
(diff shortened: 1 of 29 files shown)


Nicolapps:
* **#580** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/580?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/580?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#570** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/570?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#505** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/505?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `reboot`


<h2></h2>

packages/core/src/components/passkey/authentication.ts
```diff
@@ -126,26 +128,30 @@ export const finishAuthentication = mutation({
...
+    response: vAuthenticationResponseJSON,
   },
   returns: finishAuthenticationResult,
   handler: async (ctx, args): Promise<FinishAuthenticationResult> => {
     validatePurpose(args.purpose);
+    // `verifyAuthenticationResponse` checks that `id` and `rawId` agree, so
+    // either one can drive the lookup.
+    const credentialId = toArrayBuffer(
+      isoBase64URL.toBuffer(args.response.rawId),
+    );
```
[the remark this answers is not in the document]
Nicolapps: Not a real failure mode. `vAuthenticationResponseJSON` already pins `rawId` to `v.string()`, and `isoBase64URL.toBuffer` does not throw on a malformed base64url *string* — `@hexagon/base64` decodes leniently:

```
"!!!not base64!!!" -> 12 bytes
""                -> 0 bytes
"AB$$"            -> 3 bytes
```

It only throws for a non-string (`undefined`/`null`), which the validator makes unreachable here. A garbage `rawId` therefore decodes to garbage bytes, misses the `by_credentialId` index, and returns `UNKNOWN_CREDENTIAL` — the right answer for a credential ID that is not registered.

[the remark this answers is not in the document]
Nicolapps: Not a real failure mode. `vAuthenticationResponseJSON` already pins `rawId` to `v.string()`, and `isoBase64URL.toBuffer` does not throw on a malformed base64url *string* — `@hexagon/base64` decodes leniently:

```
"!!!not base64!!!" -> 12 bytes
""                -> 0 bytes
"AB$$"            -> 3 bytes
```

It only throws for a non-string (`undefined`/`null`), which the validator makes unreachable here. A garbage `rawId` therefore decodes to garbage bytes, misses the `by_credentialId` index, and returns `UNKNOWN_CREDENTIAL` — the right answer for a credential ID that is not registered.

examples/react-passkey/src/routes/logIn.tsx
```diff
@@ -97,6 +97,10 @@ function errorMessage(
       // This might be caused by a misbehaving client, or by a configuration error.
       // The Convex logs contain more information about the source of the error.
       return "This passkey request could not be verified. Please try again, or contact support if the problem persists.";
+    case "PASSKEY_ALREADY_REGISTERED":
+      // The authenticator refused to make a second passkey for this
+      // account (usually via `excludeCredentials`).
+      return "You already have a passkey for this account on this device. Sign in with it instead.";
```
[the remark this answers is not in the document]
Nicolapps: Fixed: the message no longer names an account. That `errorMessage` switch also serves the sign-up path, where the username was just found free and no account exists, so the copy is now account-neutral — "This device already holds a passkey for this site. Try signing in with it instead."

packages/core/src/components/passkey/react.tsx
```diff
@@ -256,12 +249,10 @@ export function useModalCeremony(
...
       // autofill hook even when the caller hands over a different one
-      // mid-ceremony. The gates delegate to their hook's current loop at
-      // call time: when the autofill effect restarts mid-ceremony, the new
-      // loop starts unpaused and the resume below is a no-op on it
-      // (last-loop-wins).
+      // mid-ceremony. The gates act on the hook, not on one loop, so the
+      // pause holds even when the autofill effect restarts mid-ceremony.
       const autofill = autofillRef.current;
-      await autofill?.pause();
+      autofill?.pause();
```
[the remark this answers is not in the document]
Nicolapps: **(Claude, acting for @Nicolapps.)** Real, and confirmed by a regression test that fails on the previous commit.

The race: `run()` calls `autofill.pause()`, which aborts the pending conditional request, then `fn()` finishes and the `finally` resumes before the aborted `get()` promise settles. The loop then wakes with intent back on `"running"`, so it read its own abort as a foreign displacement, went `idle`, and parked for good.

Fixed in b4b6e618 by reading the fact off the `AbortSignal` instead of the intent: the hook owns that `AbortController` and nobody else aborts it, so `controller.signal.aborted` means "this hook aborted its own request" for every interleaving, whatever the intent says by then. A refresh and a pause share that branch; the top of the loop parks if a pause is still owed and otherwise mints a fresh challenge. `CEREMONY_ABORTED` on a controller we did not abort keeps the displacement meaning.

New test: `a modal ceremony that ends before its abort settles restarts the loop`.

packages/core/src/components/passkey/management/add.ts
```diff
@@ -163,14 +161,29 @@ export function verifyAddPasskey(config: UsernamePasskeyConfig) {
         config.usernameComponent.public.getUsername,
         { userId },
       );
+      if (username === null) {
+        // WebAuthn requires `user.name` and `user.displayName`, and the
+        // browser keeps them in its passkey manager for the life of the
+        // passkey. Only the app can remove a username, so this is an app
+        // state the flow cannot serve, not a user error. Throwing rolls
+        // back the registration challenge.
+        throw new Error(
```
Nicolapps, line 170: This is a small breaking change. Before this PR the `add` function allowed the `username` value it returns to be `string | null`, but in practice we always need a username when creating a passkey, so let’s fail early on the server first

packages/core/src/components/passkey/management/add.ts
```diff
@@ -104,18 +101,22 @@ export function startAddPasskey(config: UsernamePasskeyConfig) {
...
 }
 
 export function verifyAddPasskey(config: UsernamePasskeyConfig) {
   return mutationGeneric({
     args: {
-      credentialId: v.bytes(),
-      authenticatorData: v.bytes(),
-      clientDataJSON: v.bytes(),
-      signature: v.bytes(),
+      response: vAuthenticationResponseJSON,
```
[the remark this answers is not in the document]
Nicolapps: No need for backwards compatibility for now

packages/core/src/components/passkey/authentication.ts
```diff
@@ -126,26 +125,28 @@ export const finishAuthentication = mutation({
...
```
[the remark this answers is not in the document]
Nicolapps: This is incorrect, the base64 decoding doesn't throw on invalid input

packages/core/src/components/passkey/client.ts
```diff
@@ -281,33 +261,60 @@ export async function runAuthenticationCeremony(options: {
...
+      success: true,
+      response: assertionToJSON(credential as PublicKeyCredential),
+    };
   } catch (cause) {
     return { success: false, userError: foldClientError(cause) };
   }
 }
+
+/**
+ * Put an assertion credential from a `get()` call into the wire shape.
```
dowski: Is this `get()` call part of the browser side ceremony? Maybe clarify since `get()` seems kind of generic.

packages/core/src/components/passkey/flows.ts
```diff
@@ -80,12 +79,7 @@ type FinishSignUpMutation = FunctionReference<
 type FinishSignInMutation = FunctionReference<
   "mutation",
   "public",
-  {
-    credentialId: ArrayBuffer;
-    authenticatorData: ArrayBuffer;
-    clientDataJSON: ArrayBuffer;
-    signature: ArrayBuffer;
-  },
+  { response: WireAuthenticationResponse },
```
dowski, line 59: Can you come up with a better name than `response`? It's kind of odd for a response to be an input argument (usually you think of it as something else, unless it's some sort of middleware or transformer function).

packages/core/src/components/passkey/authentication.ts
```diff
@@ -126,26 +125,28 @@ export const finishAuthentication = mutation({
...
-    authenticatorData: v.bytes(),
-    clientDataJSON: v.bytes(),
-    signature: v.bytes(),
+    response: vAuthenticationResponseJSON,
   },
   returns: finishAuthenticationResult,
   handler: async (ctx, args): Promise<FinishAuthenticationResult> => {
     validatePurpose(args.purpose);
+    // `verifyAuthenticationResponse` checks that `id` and `rawId` agree, so
+    // either one can drive the lookup.
```
dowski: Why is this mentioning "either one"? Is there any reason to prefer one over another?

packages/core/src/components/passkey/validation.ts
```diff
@@ -111,6 +100,160 @@ export const credentialDescriptor = v.object({
...
+  WireRegistrationResponse,
+  OpenTransports<RegistrationResponseJSON>
+>;
+
+/**
+ * The `AuthenticationResponseJSON` that the finish mutations of an
+ * authentication ceremony take. `clientExtensionResults` must be empty
+ * because the options request no extensions.
+ */
+export const vAuthenticationResponseJSON = v.object({
```
dowski, line 237: Is this name one that is already common in the protocol? If not maybe it could be called `ClientAssertionData` or something that indicates where it came from/what it's associated with. An "authentication response" both sounds like something that could be the result of anything in the whole passkey flow and also (as I think I alluded to elsewhere) it's odd for a `response` to show up as an input argument in so many places.

Nicolapps: Yes this is a name that comes from the WebAuthn spec itself: https://w3c.github.io/webauthn/#dictdef-authenticationresponsejson

packages/core/src/components/passkey/client.ts
Nicolapps: Yeah, I changed it to `navigator.credentials.get()` everywhere in 62fc6fa for clarity

packages/core/src/components/passkey/flows.ts
Nicolapps: Thanks, I added a TODO comment in 9e1e5d1835cf07799e01e466484acf4b51fc46d0. Renaming it on the top of the stack will be easier than editing the current stack

packages/core/src/components/passkey/authentication.ts
Nicolapps: I made this comment clearer in bff1993

PR #599, opened 2026-09-03
Extract the shared OAuth callback, database, and redemption modules

erquhart:
- Move the callback, database, and redemption logic out of the `oauth` component into `src/oauth/shared` so other OAuth components can reuse it. No behavior change.
- Reject an `id_token` without a `sub` claim at the callback.


packages/core/src/oauth/shared/redemption.ts
```diff
@@ -0,0 +1,173 @@
...
+  ) => Promise<{ encryptedPayload: string } | null>;
+  /** Map what the provider attested to the account profile. */
+  profile: (payload: TicketPayload<UserInfo>) => Profile;
+}) {
+  return options.authMutation({
+    args: {
+      code: v.string(),
+      state: v.string(),
+    },
+    returns: v.union(vTokenBundle, v.null()),
```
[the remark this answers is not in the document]
erquhart: Intentionally deferred

packages/core/src/oauth/shared/db.ts
```diff
@@ -0,0 +1,180 @@
...
+ */
+type MutationCtx = GenericMutationCtx<GenericDataModel>;
+
+/** What this code requires from an authorization request document. */
+type AuthorizationRequestContract = ObjectType<
+  typeof authorizationRequestFields
+>;
+
+/** What this code requires from a ticket document. */
+type TicketContract = ObjectType<typeof ticketFields>;
```
dowski, line 46: Why `ObjectType` and not `Infer`? I'm more used to the latter but maybe there's a good reason to use `ObjectType` in this case?

packages/core/src/oauth/shared/db.ts
```diff
@@ -0,0 +1,180 @@
...
+/**
+ * Record an in-flight authorization request and return the callback URL it
+ * was recorded under. Called by the app-side `startSignIn` before it
+ * redirects the user to the provider; the provider callback later claims the
+ * request by state hash.
+ *
+ * `fields` is the component's own document shape without `callbackUrl` and
+ * `expiresAt`, which are set here.
+ */
+export async function insertAuthorizationRequest(
```
dowski: Can `fields` be anything? I'm wondering about the type safety guarantees (or lack thereof) offered here.

packages/core/src/oauth/shared/db.ts
```diff
@@ -0,0 +1,180 @@
...
+  null | { expired: true; redirectTo: string } | { expired: false; doc: Doc }
+> {
+  // The wide ctx types this as a bare document, so the shape has to be
+  // asserted. The fields are not a guess, though, because every caller has
+  // to prove its `Doc` has them. Only the `_id` table brand goes unproven,
+  // and it came from a query on that table.
+  const request = (await ctx.db
+    .query("authorizationRequests")
+    .withIndex("stateHash", (q) => q.eq("stateHash", stateHash))
+    .unique()) as (Doc & { _id: GenericId<"authorizationRequests"> }) | null;
```
dowski, line 112: If each component had its own code to do these DB ops, would they be more typesafe? I wonder about the value of the shared code if it relaxes typesafety too much.

packages/core/src/oauth/shared/db.ts
erquhart: ticketFields isn't actually a validator, it's a plain object that is spread into the schema of a given provider. Infer would require it to be wrapped in an object validator.

packages/core/src/oauth/shared/db.ts
erquhart: The value of the shared code is avoiding small differences leading to security issues, and these kind of ops are sort of a hotspot for that (our recent CVE is a good example). In this case, we do know that the returned doc type is accurate because we're querying the authorizationRequests table, so its a pretty honest cast. The only way its inaccurate is through some difference in an individual component's schema, which is the same trade we'd make if we rewrote all of these functions for each component.

packages/core/src/oauth/shared/db.ts
erquhart: Yeah this should be a generic like the claim function, will update

packages/core/src/oauth/shared/callback.ts
```diff
@@ -0,0 +1,417 @@
...
+  }
+
+  const config = options.exchangeConfig(authRequest);
+
+  // From here on the flow is legitimate, so failures go back to the app
+  // as a normalized error param; raw detail goes to logs only.
+  if (params.error !== null || params.code === null) {
+    console.error(
+      `OAuth callback error from provider "${config.providerName}":`,
+      params.error ?? "missing code",
```
dowski, line 351: Is this actually an app-level error? Or should this be handled like a missing `state` value (earlier in the flow and with a `throw`)?

packages/core/src/oauth/shared/callback.ts
```diff
@@ -0,0 +1,417 @@
...
+          `Userinfo request "${key}" failed with ${response.status}: ${response.bodyText}`,
+        );
+      }
+      return [key, JSON.parse(response.bodyText) as unknown] as const;
+    }),
+  );
+  return Object.fromEntries(responses);
+}
+
+/** The parameters a provider sends to the callback. */
```
dowski, line 235: Would it help to type some of them as `| undefined` to distinguish what's expected to be missing sometimes vs. what's errorneously missing?

packages/core/src/oauth/shared/redemption.ts
```diff
@@ -0,0 +1,173 @@
...
+ * Returns null when the code is unknown, already redeemed, expired, or the
+ * state doesn't match: all indistinguishable to the caller.
+ *
+ * The component calls are subtransactions of this mutation, so a failure
+ * anywhere (including the app rejecting the sign-in from `createUser` /
+ * `onSignIn`) rolls back the ticket claim. Only a successful redemption
+ * consumes the ticket.
+ */
+// TODO: dowski - return the shared `vSignInSuccess` envelope like the other
+// providers do, instead of a bare bundle or null.
```
dowski: I think when you merge with upstream and resolve conflicts, this will be gone.

dowski (approved):

packages/core/src/oauth/shared/callback.ts
erquhart: This is the basic case where the provider returns an error instead of a code. For example if the user cancelled sign in from the provider sign in, we'd get the `"access_denied"` error here. Any other error normalizes to `"oauth_error"`. But yeah in those cases we should redirect the user back to the app with one of those two errors.

erquhart: To be clear a user could end up here due to non-app errors, but not strictly non-app errors, and there's no great way to differentiate across providers. But even if we could, we'd still want to redirect instead of stranding them on convex.site (which missing state does because we literally have no info on the flow). Logging for the developer and giving an opaque error on the redirect query params is the correct way to handle it.

dowski: Got it, that all makes sense. Thanks Shawn!

packages/core/src/oauth/shared/callback.ts
erquhart: The goal here is for individual provider components not to think about what's in the callback params at all - this is a transport type so they just pass the params and move on. If we have them start interpreting params and worrying about proper/improper values, each provider component has a chance to get something wrong.

You can't tell any of this because the branch that implements is higher in the stack, but this is how it looks for a provider component to pass callback params through to this function:

```ts
  return await runCallback<GithubRequest>({
    path: url.pathname,
    params: {
      state: url.searchParams.get("state"),
      code: url.searchParams.get("code"),
      error: url.searchParams.get("error"),
      errorDescription: url.searchParams.get("error_description"),
    },
    ...
```

dowski: Okay, I get the reasoning but it just seems odd for the contract of the related function to so readily appear to take values like `state` and `code` as `null` and then turn around and throw or error if that's the case. 

Maybe just explain in the doc why the type accepts such loosey-goosey values and that not all combinations represent valid data.

erquhart: Forgot to commit this change, it went into the next branch in the stack.

PR #602, opened 2026-09-07
Add the Apple OAuth provider

erquhart:
- Add Sign in with Apple as its own component, with `setupApple` and `useSignInWithApple`.
- Handle Apple's form_post callback, the per-exchange signed client secret, and the one-time `user` name payload.
- Add the `react-apple` example and the Apple docs section.


examples/react-apple/README.md
```diff
@@ -0,0 +1,16 @@
...
+- The Apple OAuth component ("Sign in with Apple").
+
+## Generate code / run
+
+Run the example from its own directory.
+
+```bash
+cd examples/react-apple
+npx convex dev --once    # provisions a deployment, generates convex/_generated
+npx @convex-dev/auth     # sets AUTH_PRIVATE_KEY + AUTH_JWKS on the deployment
```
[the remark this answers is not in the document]
erquhart: Precedent is to just have those general steps in the readme and actual guidance goes in docs.

examples/react-apple/vite.config.ts
```diff
@@ -0,0 +1,12 @@
+import react from "@vitejs/plugin-react";
+import { defineConfig } from "vite";
+
+export default defineConfig({
+  plugins: [react()],
+  server: {
+    // The redirect origins in convex/auth.ts point at this port.
+    // Fail instead of silently moving to another port.
+    port: 5173,
```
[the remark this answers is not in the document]
erquhart: This is probably fine and def not a p2.

packages/core/src/oauth/apple/provider.ts
```diff
@@ -0,0 +1,87 @@
...
+    };
+  },
+});
+
+/** Store a one-time redeemable ticket after a successful code exchange. */
+export const createTicket = internalMutation({
+  args: {
+    stateHash: v.string(),
+    ticketCodeHash: v.string(),
+    encryptedPayload: v.string(),
```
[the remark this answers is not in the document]
erquhart: you're conflating this with a concern about unsafe arbitrary input, doesn't apply here

packages/core/src/oauth/apple/user.ts
```diff
@@ -0,0 +1,77 @@
...
+ */
+export function sanitizeAppleUser(
+  raw: string | null,
+): SanitizedAppleUser | undefined {
+  if (raw === null) {
+    return undefined;
+  }
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(raw);
```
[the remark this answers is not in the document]
erquhart: already been fully read and body parsed by the time we get here, and size is platform bounded. not worth doing and not a p2.

erquhart:
@greptileai review

examples/react-apple/convex/schema.ts
```diff
@@ -0,0 +1,13 @@
...
+import { v } from "convex/values";
+
+export default defineSchema({
+  users: defineTable({
+    /**
+     * Apple sends the name only on the very first authorization, so it has
+     * to be stored to be used reliably.
+     */
+    name: v.optional(v.string()),
+    email: v.optional(v.string()),
```
dowski, line 11: Can these be non-optional in the schema?

packages/core/src/oauth/apple/constants.ts
```diff
@@ -0,0 +1,31 @@
...
+/** The `iss` claim Apple's id_tokens carry. */
+export const ISSUER = "https://appleid.apple.com";
+
+/**
+ * What the sign-in asks for. Apple puts the email in the id_token whenever
+ * the email scope was granted, and relays the name through the browser only
+ * on the first authorization. Requesting either scope is also what makes
+ * Apple post the callback rather than redirect to it.
+ */
+export const SCOPES = ["name", "email"];
```
dowski: Consider `readonly` and/or `as const`.

packages/core/src/oauth/apple/schema.ts
dowski, line 1: I know our goal is to put the OAuth providers into separate components, but I wonder if there's some truly tiny shared bit that should be in a common component and installed by each. I'm thinking of exactly this sort of authorization request and ticket tracking stuff.

packages/core/src/oauth/apple/secret.ts
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Put an Apple `.p8` key into the exact PKCS#8 PEM shape `importPKCS8`
+ * insists on: the header line, then the base64, then the footer line. The
+ * file Apple gives you is already in that shape, but it rarely survives the
+ * trip into an environment variable intact, so three spellings all come out
+ * the same here:
+ *
+ * - the file's contents, unchanged
+ * - the same thing with its newlines written as literal `\n`
+ * - the base64 body alone, with the header and footer lines stripped off
```
dowski: Does this "three spellings all come out the same" mean that the input `privateKey` can be formatted in any of these ways? Maybe state that directly if so.

packages/core/src/oauth/apple/secret.ts
```diff
@@ -0,0 +1,95 @@
...
+
+/** Who the client secret is addressed to, per Apple's docs. */
+const AUDIENCE = "https://appleid.apple.com";
+
+/**
+ * How long a minted secret stays valid. Apple's ceiling is six months, but
+ * the secret is signed per exchange and used immediately, so it only needs to
+ * outlive one request.
+ */
+const TTL_SECONDS = 5 * 60;
```
dowski, line 23: Generating the secret dynamically is great!

packages/core/src/oauth/apple/user.ts
```diff
@@ -0,0 +1,77 @@
...
+ * into the system sign-in sheet, so the browser relays it to the callback
+ * instead, and it arrives only on the very first authorization. Treat it as
+ * user-controlled input, which is what Apple's docs say to do.
+ *
+ * @module
+ */
+
+/**
+ * Longest first or last name kept. Anything longer is dropped as we
+ * assume this is not an actual name and shenanigans are afoot.
```
dowski, line 15: Little `Bobb${"y".repeat(256)} Longstring`

packages/core/src/oauth/apple/user.ts
```diff
@@ -0,0 +1,77 @@
...
+ */
+const MAX_NAME_LENGTH = 256;
+
+/** What survives sanitizing: a name, and never an email address. */
```
dowski: nit: don't document things that aren't present.

packages/docs/content/docs/login-providers/oauth.mdx
```diff
@@ -90,7 +91,142 @@ Pick the provider you want to wire up:
...
+
+Store all four values as [environment
+variables](https://docs.convex.dev/production/environment-variables#declaring)
+on your Convex deployment. The private key is multi-line, so read it from the
+file rather than passing it as an argument:
+
+```sh
+npx convex env set AUTH_APPLE_CLIENT_ID <your-services-id>
+npx convex env set AUTH_APPLE_TEAM_ID <your-team-id>
+npx convex env set AUTH_APPLE_KEY_ID <your-key-id>
```
dowski: nit: use something other than angle brackets to denote that these should be replaced, since the line below uses the angle bracket for file input.

packages/docs/content/docs/index.mdx
```diff
@@ -48,7 +48,7 @@ Currently Convex Auth supports several login methods out of the box:
 
 - [Password](/login-providers/password)
 - [Passkey](/login-providers/passkey)
-- [OAuth](/login-providers/oauth) (Google and GitHub)
+- [OAuth](/login-providers/oauth) (Apple, Google, and GitHub)
```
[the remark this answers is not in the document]
dowski: Yeah it's worth updating that paragraph above.

examples/react-apple/convex/schema.ts
erquhart: We could make them that way for the example, but it wouldn't be a good general recommendation, so I'd leave them.

- name only comes on the first sign in, so an app using apple sign in already that migrates to us may or may not have a name to use here. really developer's choice.
- email is actually optional for apple in some cases, like managed school accounts, so if someone with such an account tries to sign in they're blocked if email is required.

packages/core/src/oauth/apple/schema.ts
erquhart: I think we talked this one through on shared code being the simpler approach, let me know if you have more thoughts on it.

packages/core/src/oauth/apple/user.ts
erquhart: The thinking in leaving this was clarifying that email is stripped, but we can just drop it.

examples/react-apple/convex/convex.config.ts
```diff
@@ -0,0 +1,50 @@
...
+//
+// The full return URL will be something like:
+//
+// https://happy-animal-123.convex.site/oauth/apple/callback
+//
+// Apple rejects localhost and IP addresses as return URLs, so this deployment
+// needs a publicly reachable host like the convex.site url above. Local or
+// self-hosted deployments may need to take extra steps for this, e.g., using a
+// tunnel.
+// The app itself can stay on localhost, since Apple never sees where the flow
```
dowski: nit: maybe say "app frontend".

packages/core/src/oauth/apple/constants.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * Everything about Apple that never varies between apps. This component
+ * serves Apple only, so these are constants rather than configuration.
```
dowski, line 3: nit: end this with `@module` since its documenting the file.

packages/core/src/oauth/apple/provider.ts
```diff
@@ -0,0 +1,90 @@
+import { v } from "convex/values";
+import { env, internalMutation, mutation } from "./_generated/server.ts";
+import type { Doc } from "./_generated/dataModel.ts";
+import * as db from "../shared/db.ts";
```
dowski: What would you think about calling this `dbHelpers`? Calling it `db` directly feels a little confusing with the very well established `ctx.db`type.

packages/core/src/oauth/apple/provider.ts
```diff
@@ -0,0 +1,90 @@
...
+    stateHash: v.string(),
+    ticketCodeHash: v.string(),
+    encryptedPayload: v.string(),
+  },
+  returns: v.null(),
+  handler: async (ctx, args) =>
+    await db.insertTicket<Doc<"tickets">>(ctx, args),
+});
+
+/** Claim a ticket by hash, presenting the state minted at sign-in. */
```
dowski: What's "the state minted at sign-in" here? That's what the caller of this uses to compute the `stateHash`, right? It's kind of confusing to mention that here.

packages/docs/content/docs/login-providers/oauth.mdx
```diff
@@ -90,7 +91,142 @@ Pick the provider you want to wire up:
...
+
+**Register the Services ID, the key, and their environment variables**
+
+Sign in with Apple is set up in [Certificates, Identifiers &
+Profiles](https://developer.apple.com/account/resources/), and needs a paid
+Apple Developer Program membership.
+
+<Callout type="info">
+The callback lives on your deployment's `convex.site` host rather than on your
+app, so a cloud deployment works as it is and your app can stay on
```
dowski: nit: "app frontend" here too.

dowski (approved):

packages/core/src/oauth/apple/provider.ts
erquhart: Yeah agree, will improve

PR #614, opened 2026-09-10
Update the return shape for sign-in

dowski:
It now carries a status "complete" | "error" in preparation for a future where we have an "incomplete" status that needs to be resolved before it eventually completes.

Also closes a gap where the OAuth provider didn't conform to the standard sign-in success shape.

### DIFF
--- a/packages/core/src/lib/types.ts
+++ b/packages/core/src/lib/types.ts
@@ -1,6 +1,6 @@
 import { FunctionReference } from "convex/server";
 
-import { GenericId, Infer, v } from "convex/values";
+import { GenericId, Infer, v, type Validator } from "convex/values";
 
 /**
  * Shared contracts that cross a module boundary within Convex Auth. That includes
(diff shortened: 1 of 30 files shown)


dowski:
* **#614** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/614?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/614?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `reboot`


<h2></h2>

Nicolapps:
Can you please ping me for a review when you’ve addressed the Cubic review comments?

packages/core/src/oauth/client.ts
```diff
@@ -38,7 +39,7 @@ export type OauthProviderApi = {
     "mutation",
     "public",
     { code: string; state: string },
-    TokenBundle | null
+    ClientView<CompleteSignInResult>
```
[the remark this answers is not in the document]
dowski: We're not targeting backward compat right now, and also client and backend are expected to be upgraded at the same time.

packages/core/src/lib/types.ts
```diff
@@ -31,20 +31,58 @@ export const vTokenBundle = v.object({
...
  * Providers compose this into their result union rather than declaring the
- * success arm themselves. Fixing where the minted bundle sits is what lets the
+ * complete arm themselves. Fixing where the minted bundle sits is what lets the
  * SSR auth proxy find the refresh token without knowing which provider produced
  * the response, and lets it reject a shape it doesn't recognize instead of
  * forwarding tokens to the browser.
  */
-export const vSignInSuccess = v.object({
-  success: v.literal(true),
+export const vSignInComplete = v.object({
```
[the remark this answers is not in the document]
dowski: Backward compat is not a goal right now.

packages/core/src/server/signInProxy.ts
```diff
@@ -131,24 +132,42 @@ function isEncodedTokenBundle(value: unknown): value is TokenBundle {
...
+  const { status, tokens } = value as Record<string, unknown>;
+  if (typeof status !== "string" || !Object.hasOwn(MINTS_SESSION, status)) {
+    return null;
+  }
+  // An arm that mints nothing has no tokens to intercept, so it goes back as
+  // the provider wrote it. On the arms that do mint, a bundle that isn't a
+  // bundle is a wiring bug rather than something to forward.
+  if (!MINTS_SESSION[status as SignInStatus]) {
+    return Object.hasOwn(value, "tokens") ? null : { kind: "forward" };
+  }
```
Nicolapps: I think this code is a little bit confusing (especially because you need to use `Object.hasOwn(MINT_SESSION, …)`. What do you think about replacing it by a `switch` where the default branch does `satisfies never` to ensure exhaustivity?

Nicolapps: A potential footgun with the current implementation is that if `MINTS_SESSION` accidentally becomes non-finite (e.g. the type becomes `string`), `MINTS_SESSION` could accidentally become non-exhaustive

Nicolapps (approved):
Sounds good

packages/core/src/server/signInProxy.ts
dowski: I agree, thanks for the suggestion. Should be better now.

PR #592, opened 2026-08-31
usernamePasskey: automatically name new passkeys after the authenticator

Nicolapps:
When registering new passkeys, we now apply a default name depending on the authenticator to make them easier to distinguish.

PR #616, opened 2026-09-10
Move “default passkey name” behavior (component → provider)

Nicolapps:
I realized that the implementation of the “default passkey name” behavior introduced in #592 was done in the component instead of the provider function. This is probably too opinionated and it’s better to move it to the provider function, so that a custom implementation can change the behavior if it wants.


Nicolapps:
* **#616** <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/616?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/get-convex/convex-auth/616?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `reboot`


<h2></h2>

